PR #4, opened 2021-11-08
Brief hold left click when Ctrl+click in stash

mgleed:
Some D2R players have noticed that holding down left click longer is helpful when experiencing problems with control+leftclick not properly moving items to/from stash. I had this problem a couple of times with the bot and got "stash full" in log. I replaced the instant click with a short press/release hold.


aeon0:
Lol, cant belive this fixes the issue. Maybe you should also open a pr on the D2R repo...
thx

PR #5, opened 2021-11-08
Expanded pickit items and defined new naming convention

Sp3cialk29:


src/item_finder.py
```diff
@@ -49,6 +49,7 @@ def __init__(self):
         # load all templates
         self._templates = {}
         for filename in os.listdir('assets/items'):
+	    filename = filename.lower()
```
aeon0, line 52:
```suggestion
	          filename = filename.lower()
```

aeon0 (changes requested):
Can you also rename the rare_ring_mail back to bl__ring_mail_rare (or bl__rare_ring_mail). Same for the unique version

PR #16, opened 2021-11-12
custom discord name 

danonji:


src/utils/misc.py
```diff
@@ -7,10 +7,12 @@
 import requests
 
 
-def send_discord(item_name, url:str = None):
+def send_discord(item_name,custom_discord_name, url:str = None):
```
aeon0, line 10: Missing whitespace at `,custom_discord_name`

src/utils/misc.py
```diff
@@ -7,10 +7,12 @@
...
     if url is None:
         url = "https://discord.com/api/webhooks/908071105372250213/puaS6gIYqYxTE-TBLAIs6_Qb6ZUwuygSeQfTQkpuXrSag5DPeV2gk0SctOjPy5qMHGeh"
-    data = {"content": f"Botty just found: {item_name}"}
+    data = {"content": f"Botty just found: {item_name}"
+            "username": "{custom_discord_name}"
+           }
```
aeon0, line 15: Please format nicely. One line is good enough here.
Also, how do you like the idea to type "Botty" in case no custom_discord_name is selected. And instead replace it with the custom name?

aeon0 (changes requested):
Hi, thanks for the PR. Nice touch. But needs some rework, check my comments : )

aeon0:
There was another PR that introduced this feature.

PR #17, opened 2021-11-12
implemented loot_screenshots

Cr4zyEight:
Hi aeon0
I've implemented the functionality of creating loot screenshot of every single loot the bot is scanning.
Can be useful to see if the bot is missing out some good stuff.
Please review the code as it's my first python code i ever wrote so it was more like seeing what u did so far and adapting it.


src/pickit.py
```diff
@@ -27,6 +28,11 @@ def pick_up_items(self, char: IChar) -> bool:
         found_items = False
         keyboard.send(self._config.char["show_items"], do_press=True, do_release=False)
         time.sleep(1.0) # sleep needed here to give d2r time to display items on screen on keypress
+        #Creating a screenshot of the current loot
+        if self._config.general["loot_screenshots"]:
+            img = self._screen.grab()
+            cv2.imwrite("./loot_screenshots/info_debug_drop_" + str(time.time()) + ".png", img)
```
aeon0, line 34: Ah see, there you added a timestamp. I would suggest having a more readable one e.g. `time.strftime('%X-%x')`. This way it is easier to cross-check with the log.

aeon0 (changes requested):
GJ. Seems like you will be a python master in no time : )

tical666:
Don't forget to add the directory to .gitignore

src/pickit.py
Cr4zyEight: yes thats my thoght also but didnt found out so fast how to add better timestamp :D

Cr4zyEight:
Ok added timestamp and also a Readme entry (same as in info_screenshots branch :D)

aeon0 (approved):

PR #18, opened 2021-11-12
botty folder not so cluttered

Cr4zyEight:
All screenshots like 
  info_debug_chicken*
  info_debug_drop*
  info_wait_for*
go to a separate folder "info_screenshots"
So botty folder not getting confusing


aeon0:
Hi, good idea. 2 improvements I see. 
1) Can you add a timestamp to each filename (as it is in the .log file). Then it will not override the previous as it currently does
2) Did you test if cv2. imwrite will create the folder if it does not exist, or will it have a runtime error?

mgleed:
will also need to update .gitignore when this is done

Cr4zyEight:
> Hi, good idea. 2 improvements I see.
> 
> 1. Can you add a timestamp to each filename (as it is in the .log file). Then it will not override the previous as it currently does
> 2. Did you test if cv2. imwrite will create the folder if it does not exist, or will it have a runtime error?

1 -> isnt there a timestamp for each file? when i run bot for example 1 hour i have like 60 screenshots in my folder
2 -> it will not create the folder by itself thats why i added the info.txt to the folder so its getting commited with git

aeon0:
I hope you will not have 60 screenshots :D But yeah, could happen. But then it is easier to find the real rootcause of issus.

Cr4zyEight:
- Ok i added a time stamp that will look like this "info_debug_chicken_20211115_230959.png" so it will be easy to order it by time in the folder.

- Also i added the png files in the folder "info_screenshots" to gitignore... there is already a ignore entry that excludes all png files but two is better than one and this one especially goes for the folder ;P 

- And its now in the params.ini so it can be turned off (default is currently on)

- Also added Readme entry (clear the pngs once in a while)

My idea would be to implement something generally that clears off png files that are older than lets say 1 or 2 weeks to not getting spammed by those guys who dont read the readme "my 6 TB Volume is full only by botty XD"
But let me hear your opinion :P

PR #19, opened 2021-11-13
Support 1280x720

aeon0:


bjoerm:
Hey, will this enable the bot to work at any resolution, so that 2560x1440 is also supported? Or do you need to have all screenshots taken again for each resolution?

freddymuch:
we should add 'eth_weapon_great_poleaxe' which is the most valuable basic item to make _infinity_

aeon0:
It is included : )

PR #20, opened 2021-11-14
fix anya templates and images

reqyl:
- Add template mapping for anya
- The current images seems to have smaller resolution and anya can't be found with default configuration.


reqyl:
PS. I'm doing a javazon gg gloves shopper. If there's interest I could make a PR later on

PR #25, opened 2021-11-14
Shop for claws

aeon0:


M4ntis91:
Nice work so far.
I modified the claw.py for myself and tried to implent it as "run_anya" to the bot.py but failed. could you help me out with it? would like to implent a function to create a new shop - game after ~2hours, to protect the game from crashes.

aeon0:
@M4ntis91 sry, didnt see the comment, this pr is stale now, now PR #148 and I think the plan is to add such a feature there!

PR #28, opened 2021-11-15
actually stashing gold (wasnt implemented yet)

Cr4zyEight:
guess commit message is enough :)

### DIFF
--- a/src/ui_manager.py
+++ b/src/ui_manager.py
@@ -12,8 +12,6 @@ import numpy as np
 from logger import Logger
 from utils.misc import wait, cut_roi, color_filter, send_discord
 from config import Config
-import threading
-
 
 class UiManager():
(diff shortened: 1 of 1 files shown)


src/ui_manager.py
```diff
@@ -10,9 +10,9 @@
 import os
 import numpy as np
 from logger import Logger
-from utils.misc import wait, cut_roi
+from utils.misc import wait, cut_roi, color_filter, send_discord
 from config import Config
-from utils.misc import color_filter
+import threading
```
aeon0, line 15: why import threading here?

aeon0, line 15:
```suggestion
```

Cr4zyEight: That was for my custom discord hook to see if hes actually stashing :) just forgot to delete the import

Cr4zyEight:
deleted the unused import

PR #42, opened 2021-11-17
Belt management

mgleed:
### bot
1. Opens belt and parses contents.
2. If potions don't belong based on belt_slots param, will drop
3. Then pickit to resort pots
(I put the sort_belt function in this class because it had calls to both belt_manager and pickit which would then have dependencies on each other if I had put the function in belt_manager--I don't know how to resolve this)

### health_manager
1. added support for rejuv potions

### belt_manager
-various functions for sorting/parsing potions
-reworked corpse pickup pot filling routine

### pickit
-now filters item_list for potions that cannot be picked up with respect to order of columns or if the column is full

### template_finder
-added List as a potential argument to search(), the same behavior I added to search_and_wait() in one of my first PR's. This way I can search for rejuv or full rejuv with one call to search().

There are a variety of other small changes.

In params.ini I changed `take_health_potion` to `take_healing_potion`. I realize this will be a bit of a pain for users who already have custom.ini, but the healing_potion is a match for variables in the program I wrote including for template files like super_{potion_type}. If this is too big of a deal, the object property name could probably be renamed while the property value could be kept the same in config

Looking forward to your thoughts.

Will update readme if we decide to move forward


mgleed:
One to-do. I need to adjust fill_from_inventory to handle cases where there are no remaining pots in more than one column (i.e. if death and no HP potions were in # slot, then any other remaining empty slots to the right can't be filled unless they are dragged to the slot manually)

src/belt_manager.py
```diff
@@ -0,0 +1,205 @@
...
+                    break
+                wait(0.1, 0.2)
+        wait(0.5, 0.65)
+        keyboard.release("shift")
+        keyboard.send(self._config.char["inventory_screen"])
+
+
+if __name__ == "__main__":
+    keyboard.wait("f11")
+    #idk what this does
```
aeon0, line 205: This part wil be called in case you directly execute `python src/belt_manager.py`. I usually use this for testing.
This code will not be called if you do `from belt_manager import BeltManager`

src/bot.py
```diff
@@ -30,18 +31,19 @@ def __init__(self):
         self._item_finder = ItemFinder()
         self._ui_manager = UiManager(self._screen, self._template_finder)
         self._pather = Pather(self._screen, self._template_finder)
-        self._health_manager = HealthManager(self._screen, self._template_finder, self._ui_manager)
+        self._belt_manager = BeltManager(self._screen, self._template_finder)
+        self._health_manager = HealthManager(self._screen, self._template_finder, self._ui_manager, self._belt_manager)
+        self._pickit = PickIt(self._screen, self._item_finder, self._ui_manager,self._belt_manager)
```
aeon0, line 36: missing whitespace

src/bot.py
```diff
@@ -77,6 +79,17 @@ def draw_graph(self):
...
+        # here instead because pickit depends on belt_manager and belt_manager depends on pickit
+        wait(0.4,0.6)
+        keyboard.send(self._config.char["show_belt"]) #toggle belt
+        wait(0.1,0.2)
+        img = self._screen.grab()
+        belt_contents, self._belt_manager.belt_height = self._belt_manager.get_belt_contents(img,1,0)
+        self._belt_manager.drop_wrong_belt_pots(belt_contents,0) #drops wrong potions and generates pots_remaining
+        _ = self._pickit.pick_up_items(self._char) #pickup compatible pots
+        keyboard.send(self._config.char["show_belt"]) #toggle belt
+
```
aeon0, line 92: This should be part of the belt manager. Why does BeltManager depend on Pickit? I don't really get the pickit call after drop wrong belt pots. Do you drop them and then pick up again to make sure they are in the right order?

src/bot.py
```diff
@@ -117,6 +130,7 @@ def on_create_game(self):
         # Make sure these keys are released
         keyboard.release(self._config.char["stand_still"])
         keyboard.release(self._config.char["show_items"])
+
```
aeon0, line 133: remove empty line here

src/health_manager.py
```diff
@@ -10,44 +10,45 @@
...
+                    Logger.debug(f"Give {potion_type} in slot {i+1} to merc, val={(percentage*100):.1f}%")
                     keyboard.send(f"left shift + {self._config.char[key]}")
                 else:
-                    Logger.debug(f"Drink {potion_type} potion in slot {i+1}")
+                    Logger.debug(f"Drink {potion_type} in slot {i+1}, val={(percentage*100):.1f}%")
                     keyboard.send(self._config.char[key])
+                self._belt_manager.potions_remaining[i] = self._belt_manager.potions_remaining[i] - 1
                 break
+        # if not key:
+        #     Logger.debug(f"Out of {potion_type}")
```
aeon0, line 51: commented code. uncomment or delete

src/health_manager.py
```diff
@@ -100,32 +101,45 @@ def start_monitor(self, run_thread: Thread):
...
                 health_percentage = self.get_health(img)
+                last_rejuv_drink = time.time() - self._last_rejuv
                 last_drink = time.time() - self._last_health
-                if health_percentage < self._config.char["take_health_potion"] and last_drink > 3.5:
-                    self._drink_poition(img, "health")
-                    self._last_health = time.time()
+
                 # give the chicken a 6 sec delay to give time for a healing pot and avoid endless loop of chicken
-                elif health_percentage < self._config.char["chicken"] and (time.time() - start) > 6:
+                if health_percentage < self._config.char["chicken"] and (time.time() - start) > 6:
```
aeon0, line 108: We have higher prio on chicken vs rejuv potion. Shouldnt it be the other way around. Bascially, try to take rejuv potion, only if you don't have anymore, do chicken?

src/health_manager.py
```diff
@@ -100,32 +101,45 @@ def start_monitor(self, run_thread: Thread):
...
+                    self._last_rejuv = time.time()
+                elif health_percentage < self._config.char["take_healing_potion"] and last_drink > 3.5 and last_rejuv_drink > 1.5:
+                    self._drink_potion("healing_potion",health_percentage)
+                    self._last_health = time.time()
                 # check mana
                 mana_percentage = self.get_mana(img)
                 last_drink = time.time() - self._last_mana
-                if mana_percentage < self._config.char["take_mana_potion"] and last_drink > 4:
-                    self._drink_poition(img, "mana")
+                last_rejuv_drink = time.time() - self._last_rejuv
```
aeon0, line 121: don't think we have to recalculate that here. do we?

src/health_manager.py
```diff
@@ -100,32 +101,45 @@ def start_monitor(self, run_thread: Thread):
...
+                elif mana_percentage < self._config.char["take_mana_potion"] and last_drink > 4 and last_rejuv_drink > 2:
+                    self._drink_potion("mana_potion",mana_percentage)
                     self._last_mana = time.time()
                 # check merc
                 merc_alive, _ = self._template_finder.search("MERC", img, roi=self._config.ui_roi["merc_icon"])
                 if merc_alive:
                     merc_health_percentage = self.get_merc_health(img)
                     last_drink = time.time() - self._last_merc_healh
+                    last_merc_rejuv_drink = time.time() - self._last_merc_rejuv
                     if merc_health_percentage < self._config.char["merc_chicken"]:
```
aeon0, line 134: For merc we could leave chicken this way, cause there is a delay in giving rejuv and merc consuming it. While for own char, there is no delay

src/pickit.py
```diff
@@ -47,28 +52,36 @@ def pick_up_items(self, char: IChar) -> bool:
...
+                    # no need to stash potions, scrolls, or gold
                     if (("potion" not in closest_item.name) and ("tp_scroll" != closest_item.name) and ("misc_gold" not in closest_item.name)):
                         found_items = True
+
+                    #if picking up potion, then add to potions_remaining
+                    if ("potion" in closest_item.name):
+                        predictPotCol=self._belt_manager.predict_pot_col(closest_item.name)
+                        #Logger.debug(f"predictPotCol={predictPotCol}")
+                        if predictPotCol >= 0:
+                            self._belt_manager.potions_remaining[predictPotCol] = self._belt_manager.potions_remaining[predictPotCol] + 1
```
aeon0, line 64: You have to be a bit careful here. Pickit might "see" the same potion multiple times.

src/template_finder.py
```diff
@@ -92,8 +92,15 @@ def __init__(self, screen: Screen, scale_factor: float = 0.7):
...
-            "SUPER_HEALING_POTION": [load_template("assets/templates/super_healing_potion.png", 1.0), 1.0],
-            "SUPER_MANA_POTION": [load_template("assets/templates/super_mana_potion.png", 1.0), 1.0],
+            "INV_SUPER_HEALING_POTION": [load_template("assets/templates/inv_super_healing_potion.png", 1.0), 1.0],
+            "INV_SUPER_MANA_POTION": [load_template("assets/templates/inv_super_mana_potion.png", 1.0), 1.0],
+            "INV_FULL_REJUVENATION_POTION": [load_template("assets/templates/inv_full_rejuvenation_potion.png", 1.0), 1.0],
+            "INV_REJUVENATION_POTION": [load_template("assets/templates/inv_rejuvenation_potion.png", 1.0), 1.0],
+            "BELT_SUPER_HEALING_POTION": [load_template("assets/templates/belt_super_healing_potion.png", 1.0), 1.0],
+            "BELT_SUPER_MANA_POTION": [load_template("assets/templates/belt_super_mana_potion.png", 1.0), 1.0],
+            "BELT_FULL_REJUVENATION_POTION": [load_template("assets/templates/belt_full_rejuvenation_potion.png", 1.0), 1.0],
+            "BELT_REJUVENATION_POTION": [load_template("assets/templates/belt_rejuvenation_potion.png", 1.0), 1.0],
```
aeon0, line 102: To me belt and inv look the same, why we have two different ones?

src/template_finder.py
```diff
@@ -161,26 +168,32 @@ def search(self, ref: Union[str, np.ndarray], inp_img: np.ndarray, threshold: fl
...
+            rh *= scale
+
+            if img.shape[0] > template.shape[0] and img.shape[1] > template.shape[1]:
+                self.last_res = cv2.matchTemplate(img, template, cv2.TM_CCOEFF_NORMED)
+                _, max_val, _, max_pos = cv2.minMaxLoc(self.last_res)
+                self.last_score = max_val
+                if max_val > threshold:
+                    ref_point = (max_pos[0] + int(template.shape[1] * 0.5) + rx, max_pos[1] + int(template.shape[0] * 0.5) + ry)
+                    ref_point = (int(ref_point[0] * (1.0 / scale)), int(ref_point[1] * (1.0 / scale)))
+                    return True, ref_point
```
aeon0, line 196: this changes seem to be more template finder specific, but are not 100% needed for the belt manager. Since the belt manager is quite a big change, lets keep it to it and split this change into another PR?

ui.ini
```diff
@@ -53,10 +53,10 @@ mana_top=915
...
 mana_width=10
 mana_height=151
-potion1_x=1113
-potion1_y=1042
-potion_width=45
-potion_height=45
+potion1_x=1084
+potion1_y=844
+potion_width=62
+potion_height=56
```
aeon0, line 59: Where my previous coordinates so much off?

src/belt_manager.py
```diff
@@ -0,0 +1,205 @@
...
+            wait(0.2,0.3)
+            keyboard.send(self._config.char["show_belt"]) #toggle belt off
+            wait(0.2,0.3)
+
+                #if belt_contents[x][y] is not self._config.char
+
+    def fill_up_belt_from_inventory(self):
+        """
+        Fill up your belt with pots from the inventory e.g. after death. It will open and close invetory by itself!
+        Updated 11/2021 to respect assigned potion columns
```
aeon0, line 162: No need to put any dates or anything about changes. That's why we have source control, and commit messages : )

aeon0 (changes requested):
Overall I think we overcomplicate that whole thing a bit here. I like the idea of #42 where we just say we want x columns of this type of potions and add also rejuv to this.
The "only" thing missing would be to check for whole belt and not just bottom row.

src/belt_manager.py
mgleed: Thank you for your comments! Will close the PR for now based on our discord discussion.

mgleed:
Thank you for your comments! Will close the PR for now based on our discord discussion.

PR #45, opened 2021-11-17
Require mercenary

mgleed:
Added param for require_merc

If require_merc set to 1, will attempt two revives of merc. If can't revive on first attempt, will end run and retry one more time. If no luck will shut down bot.


src/bot.py
```diff
@@ -188,14 +193,33 @@ def on_maintenance(self):
...
+        if merc_alive:
+            self._merc_revive_attempts = 0
+        elif self._config.general["require_merc"]:
+            self._merc_revive_attempts = self._merc_revive_attempts + 1
+            mouse.move(0,0, randomize=[5, 5], delay_factor=[0.5,1])
+            wait(0.2, 0.3)
+            mouse.click(button="left")
+            wait(0.3, 0.5)
+            mouse.click(button="left")
+            wait(0.5, 1)
```
aeon0, line 214: What is this supposed to do?

aeon0 (changes requested):
I wouldnt make it so complicated. Have the "require_merc" param and check merc_alive after resurrection. Then just:
```python
if not merc_alive and self._config.general["require_merc"]:
  Logger.error("Merc is required but can not be resurrected")
  shut_down_d2()
  os._exit(1)
```

src/bot.py
mgleed: @aeon0 For one, sometimes resurrects are not performed because the NPC can't be located and times out, so then character will attempt run. If that case happens, the way you have written it will cause it to close out rather than try one more time.

The reason for that particular code block is that if you don't have enough gold to resurrect the merc, you need to click away twice to close the "not enough gold" dialogue.

aeon0: But doesnt it then vialoate the "require_merc" when you run once without. Seems a little hacky to me to say, "ok we try it one more time next round". And bloats up the code.
What about increasing the time_out to look for Qual if require_merc = True?

Also side note. Lets try not to have any implementation details in bot.py. It already gets quite big. A good indicator that code should not be inside bot.py is if it uses the mouse and keyboard module.

mgleed: I've watched my bot endlessly chase NPC's maybe twice before ultimately timing out and my mercless bot runs runs into Pindle's warm embrace, so that's why I implemented it that way. :P But hopefully more time should be sufficient.

src/npc_manager.py
```diff
@@ -58,7 +58,11 @@ def __init__(self, screen: Screen, template_finder: TemplateFinder):
     def open_npc_menu(self, npc_key: Npc) -> bool:
         roi = self._config.ui_roi["cut_skill_bar"]
         start = time.time()
-        while (time.time() - start) < 35:
+        if self._config.general["require_merc"]:
+            timeout = 70
+        else:
+            timeout = 35
```
aeon0, line 64: This can be one line:
```python
timeout = 70 if self._config.general["require_merc"] else 35
```

aeon0 (changes requested):
Maybe merge the latest code. there is no more shut down function. Instead we will try not to shut down anything. Just use the logger to log a message, send something to discord and close botty. But not d2r.

mgleed:
@aeon0 could I request the shut down function be kept? There is no reason here to leave d2 open consuming power. I realize this was changed because someone lost an item when they ran into a stash full bug, but I think the shut down function is nice personally.

jagarop:
> @aeon0 could I request the shut down function be kept? There is no reason here or if a stash is full to leave d2 open consuming power. I realize this was changed because at least one person lost an item when they ran into a stash full bug, but I think the shut down function is nice personally, and would be worth keeping for at least this case with no Merc.

we could bring back src/utils/misc.py/shut_down_d2() and move all unrecoverable errors that stop bot to that function. After that we can add a option in params.ini to chose to taskkill/os.exit or just os.exit.

aeon0:
Okay, how about this. Whenever there is no reason to keep d2 alive we first call the game recovery which should send you to the hero selection screen quite reliably. Only then wen shut down d2. 

So a shut_down function would look like this:
```python
game_recovery.go_to_hero_selection()
Logger.log("...")
if discord:
  # do discord stuff
kill d2 process (and  bnet process)
os._exit(1) # shut down botty itself
```

PR #47, opened 2021-11-17
Allow to control desired hp/mp column counts

reqyl:


reqyl:
Oh didn't see #42 PR before I did this.

I closing this in favor of #42 since it seems more thorough.

PR #48, opened 2021-11-18
add advanced params section with pathing delay

wittekin:
Adds a new params section, "advanced_options"
Adds a new param to the advanced_options section, "pathing_delay_factor"

The pathing delay factor will allow users to scale the otherwise static delays in the pathing `move` operations based on their own needs.


aeon0 (approved):
thx for the update!

PR #61, opened 2021-11-19
Added requirements.txt

bjoerm:


aeon0:
Why? There is a environment.yml. I dont see the need for this.

bjoerm:
Didn't notice. Then we don't need to have this.

PR #62, opened 2021-11-19
Added small explanation to JavaShopper class.

bjoerm:


karibuTW:
Thanks for this, but in this case, if we follow the steps, it will just say:

File "C:\Users\ADMIN\Downloads\botty-master\src\shop\ias_gloves.py", line 6, in <module>
    from screen import Screen
ModuleNotFoundError: No module named 'screen'

Should we explain how to load all this?

bjoerm:
Feel free to add any steps you miss. For me it works as described above.

karibuTW:
Okay, let me try later to load the modules and explain if any missing steps for new comers. Thanks

karibuTW:
I guess I am missing to load the screen setup or something like this.
You have configured the init?

I'm not sure how to make it run then

Thanks

PR #64, opened 2021-11-19
Named bot

cpamabotte:
Add the possibility to name the bot in terminal and discord.

When using several bots, it is convenient to know which one is stuck, or has dropped an item, when looking at the discord channel.
This (trivial) PR implement a very basic way to do it: the user can register in custom.ini a `name` field which is then used in `send_discord_thread`

**Example**
Setting in custom.ini
`name=My favorite bot`
will lead to message such as
> My favorite bot just found: uniq_armor_ormus_robes (v0.2.2-dev)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ run_shenk=0
 
  [general]                      | Descriptions
 --------------------------------|---------------------------------------------
+name | Name used in terminal and discord messages
 monitor | Select on which monitor D2R is running in case multiple are available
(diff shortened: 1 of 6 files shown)


cpamabotte:
I just saw the PR #16 ... But I think mine is a bit more general if in the future there is more advanced logging / different hooks / webserver interface to it, instead of being tied to discord.

aeon0:
Also I think #16 is dead. Doesnt seem like there will be any more updates. I like your approach.

PR #66, opened 2021-11-19
Allow to control desired hp/mp column

reqyl:
Reopening this PR since I'm not sure if #42 is going to be merged or not. (I couldn't reopen the existing PR since I force pushed due to rebase)

The changes in this PR are pretty small and are useful to my runs. (Never getting only hp-pots or only mana-pots)


aeon0:
Hi, I like the change. It's helpful and simple! Yet, I would probably choose to bring forward the full blown belt manager. Hope you understand.

PR #68, opened 2021-11-20
Plus 2 gloves are fine & more randomness in JavaShopper pauses

bjoerm:
- Added the option to also look for +2 skills & IAS gloves and not only for +3 skills. Option is deactivated by default and can be enabled in the JavaShopper class to keep its settings away from rest of bot.
- Also converted the fixed pauses into more random wait intervals for paranoia reasons.


reqyl:
Nice! I agree +2 is better than none 👍

src/shop/ias_gloves.py
```diff
@@ -98,7 +99,24 @@ def shop_loop(self):
...
+                            ),
+                            inp_img=img,
+                            threshold=0.80,
+                            normalize_monitor=True,
+                        )
+                        if g_gloves_found:
+                            mouse.click(button="right")
+                            Logger.info("G gloves bought!")
+                            self.gloves_bought += 1
+                            wait(1, 3)
```
aeon0, line 119: This code is redundent with L99-L102. Rather searchfor +2 gloves in case gg_gloves not found and the config is set. Then have an or for actually buying it. e.g.:
```python
if self.look_also_for_plus_2_gloves and not gg_gloves_found:
  # search for +2 gloves
if gg_gloves_found or g_gloves_found:
  # buy gloves
```
Also I like verbose variable names, but could maybe be a little shortned. At the very least I would remove the "also" from the variable name.

aeon0 (changes requested):
Thx for the PR. 

Randomness on the waits might be a bit much. I imagine it will slow down shopping quite a bit.

Sp3cialk29:
> * Added the option to also look for +2 skills & IAS gloves and not only for +3 skills. Option is deactivated by default and can be enabled in the JavaShopper class to keep its settings away from rest of bot.
> * Also converted the fixed pauses into more random wait intervals for paranoia reasons.

why not +3 | 20ias

PR #77, opened 2021-11-21
( ͡° ͜ʖ ͡°)

mgleed:
Alternatively, could replace all occurrences of botty and change to booty.


aeon0:
Ok, good for a tmp solution. But of course your proposed solution would be the end goal!

lgndz:
actually lmao - i love it

PR #96, opened 2021-11-23
support list input for template_finder.search()

mgleed:
Think this could be a helpful feature for future stuff I might work on, wanted to merge earlier than later. I wanted to also return the name of the template found, but not sure how to do that since `ref` can be an image, not just a string or list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ botty_v*/
 custom.ini
 *info_*.png
 .vscode
+.vs/
 info_screenshots/
 loot_screenshots/
 .venv
--- a/src/bot.py
+++ b/src/bot.py
@@ -259,7 +259,7 @@ class Bot:
                 portal_found = bot._char.select_by_template("A5_RED_PORTAL")
                 time.sleep(0.5)
                 pindle_start_found = bot._template_finder.search_and_wait(["PINDLE_0", "PINDLE_1"], threshold=0.65, time_out=20)
-                if not (portal_found or pindle_start_found):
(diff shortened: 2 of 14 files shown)


aeon0:
Yeah, could be useful, you could return None in case ref is an image.
Right now it will return the first template that is higher then the threshold? Do we want that, or is it better to search for all refs and return the one with the highest score?

mgleed:
Added an input parameter `best_match` for the search function, and added a third output with the template name. This is untested as I haven't come up with a test use case for it yet (I had something like this for rejuv vs. full rejuv potion in old belt manager), but I can try tomorrow. @aeon0

mgleed:
Changed the output type from (bool, float) to an object containing name, position, and score. Looked for all the places in sourcecode that used template_finder search() and search_and_wait() and adjusted accordingly. Reworked the logic on a couple spots.

Runs well. Have run a few runs, resurrected merc, and a few other things without issue but have not comprehensively tested outside of that.

Demo of best_match on vs best_match off:
search_bestMatch_off
search_bestMatch_on
 
@aeon0 let me know what you think

mgleed:
K should be done with it til review now. :p

aeon0:
Could you maybe make this branch again on the base botty repo, not on your fork. Would make life much easier for me to test stuff.
while on your branch this should do it
```
git remote add mgleed your_git_repo
git fetch mgleed
git checkout mgleed/multiSearch
# now you on a detached head
git checkout -b multiSearch
git push -u origin multiSearch
```

PR #102, opened 2021-11-24
Moves the mouse to the corner before taking a screenshot. 

bjoerm:
Prevents thus mouse from covering loot.

Here an example where this might have happened (ring got not picked up).
not picked up ring - info_debug_drop_20211122_203831

### DIFF
--- a/src/screen.py
+++ b/src/screen.py
@@ -5,6 +5,7 @@ import time
 from logger import Logger
 from typing import Tuple
 from config import Config
+from utils.custom_mouse import mouse
 import os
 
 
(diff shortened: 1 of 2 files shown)


src/screen.py
```diff
@@ -50,7 +51,10 @@ def convert_abs_to_monitor(self, abs_coord: Tuple[float, float]) -> Tuple[float,
         monitor_coord = self.convert_screen_to_monitor(screen_coord)
         return monitor_coord
 
-    def grab(self) -> np.ndarray:
+    def grab(self, hide_mouse=False) -> np.ndarray:
+        if hide_mouse:
+            mouse.move(0, 0)  # Moving the mouse to the edge of the screen to not cover any loot.
```
aeon0, line 56: You will have to convert 0, 0 to monitor coordinates. If someone is running a two monitor setup the (0, 0) screen coordinate might be at (1920, 0) for example.

aeon0 (changes requested):
Functionality is useful. But I dont like it in the screen grabbing. Maybe add a function to custom_mouse.py e.g. `mouse.hide()`. In the hide function it would be cool if it randomly moves it to any of the 4 corners with a bit of a randomized area.

bjoerm:
Thanks for the feedback. Will think about it.

bjoerm:
Didn't have time recently to work on this. Shall this be closed for now, @aeon0?

aeon0:
Stale PR

PR #107, opened 2021-11-24
update run.py for graphic debugger

mgleed:


aeon0:
changed that with the other pr #109

PR #108, opened 2021-11-24
Telekinesis for pots/gold if set

mgleed:
Tested about 10 runs with sorc, seems to work well. I don't have hammerdin to ensure that there's no issue, but I don't think there'd be any problem.

Will plan to revisit WP and stash telekinesis later, but this should be good-to-go. Ty for the pointers.


src/char/sorceress.py
```diff
@@ -19,6 +19,17 @@ def __init__(self, skill_hotkeys, char_config, screen: Screen, template_finder:
         super().__init__(skill_hotkeys, char_config, screen, template_finder, ui_manager)
         self._pather = pather
 
+    def pick_up_item(self, pos: Tuple[float, float], item_name: str = None):
+        if self._skill_hotkeys["telekinesis"] and any(x in item_name for x in ['potion', 'misc_gold']):
```
aeon0, line 23: nice syntax! : )

src/char/sorceress.py
```diff
@@ -19,6 +19,17 @@ def __init__(self, skill_hotkeys, char_config, screen: Screen, template_finder:
...
+            keyboard.send(self._skill_hotkeys["telekinesis"])
+            wait(0.1, 0.2)
+            mouse.move(pos[0], pos[1])
+            wait(0.1, 0.2)
+            mouse.click(button="right")
+            wait(0.35,0.45)
```
aeon0, line 29: utilize `self._cast_duration` here

src/char/sorceress.py
```diff
@@ -19,6 +19,17 @@ def __init__(self, skill_hotkeys, char_config, screen: Screen, template_finder:
...
+        else:
+            super().pick_up_item((pos[0], pos[1]),item_name)
```
aeon0, line 31: `(pos[0], pos[1])` is the same as `pos`, no?

aeon0 (changes requested):
Good job!
My only question is, does telek. have a range? So could it happen that it is out of range and it will just cast telek. forever??

mgleed:
> Good job! My only question is, does telek. have a range? So could it happen that it is out of range and it will just cast telek. forever??

I can work on these changes in a couple hours. The pickup item function only occurs if the item is within a certain distance,  otherwise char.move is called as you've written in pickit.py so range shouldn't be an issue.

mgleed:
I've implemented the changes you asked @aeon0 . I'm not really sure the cast_duration change was necessary since the time it takes to check all templates is so long. If you agree, I can rewind a commit and just fix the `pos` thing. The delay at the end of the pick is to ensure the item name is off the screen before doing another screen capture.

mgleed:
Overall I don't know if performance improves meaningfully with this implementation of telekinesis, but I think it's cool ;). The ideal would be to integrate item_dist; i.e., if the item is at mid-range then use telekinesis, close proximity left click, and far range tele and left click.

src/pickit.py
```diff
@@ -68,14 +68,11 @@ def pick_up_items(self, char: IChar) -> bool:
...
-                    # no need to stash potions, scrolls, or gold 
+                    # no need to stash potions, scrolls, or gold
                     if "potion" not in closest_item.name and "tp_scroll" != closest_item.name and "misc_gold" not in closest_item.name:
                         found_items = True
                     Logger.info(f"Picking up: {closest_item.name}")
-                    mouse.move(x_m, y_m)
-                    time.sleep(0.1)
-                    mouse.click(button="left")
-                    time.sleep(0.5)
+                    prev_cast_start = char.pick_up_item((x_m, y_m),item_name=closest_item.name,prev_cast_start=prev_cast_start)
```
aeon0, line 75: You really dont like whitespaces, do you? :D Please add them after `, `

aeon0 (changes requested):
Did you do a test run? I am not sure why you dont want to the cast duration from i_char which is determined by the parameters (casting_frames). Telek. should be effected by it.

I will also test this, but looks good so far.

mgleed:
> Did you do a test run? I am not sure why you dont want to the cast duration from i_char which is determined by the parameters (casting_frames). Telek. should be effected by it.
> 
> I will also test this, but looks good so far.

I included self._cast_duration in that commit for completeness sake, but nested in a conditional block. Due to pickit time it's pretty much inevitable that the delay between picks is greater than cast duration.

src/char/sorceress.py
```diff
@@ -19,6 +20,23 @@ def __init__(self, skill_hotkeys, char_config, screen: Screen, template_finder:
...
+        if self._skill_hotkeys["telekinesis"] and any(x in item_name for x in ['potion', 'misc_gold']):
+            keyboard.send(self._skill_hotkeys["telekinesis"])
+            wait(0.1, 0.2)
+            mouse.move(pos[0], pos[1])
+            wait(0.1, 0.2)
+            mouse.click(button="right")
+            # need about 0.4s delay before next capture for the item not to persist on screen
+            cast_start = time.time()
+            interval = (cast_start - prev_cast_start)
+            cast_duration_wait = (self._cast_duration - interval)
```
aeon0, line 33: you do use self._cast_duration, sry ^^

mgleed:
Fixed the white whitespace,let me know if there's anything else,thank you

src/char/sorceress.py
```diff
@@ -19,6 +20,23 @@ def __init__(self, skill_hotkeys, char_config, screen: Screen, template_finder:
...
+            delay = 0.35 if cast_duration_wait <0 else (0.35+cast_duration_wait)
+            wait(delay,delay+0.1)
+            return cast_start
+        else:
+            super().pick_up_item(pos)
```
aeon0, line 38: return here and pass through params
```suggestion
            return super().pick_up_item(pos, item_name, prev_cast_start)
```

PR #109, opened 2021-11-24
Item scores

mgleed:
Add last_item.score to item pickup as a % confidence. Suspect this will be useful for support, helping users decide on threshold, etc.


aeon0:
Nice

PR #110, opened 2021-11-24
removed hostkey press for BH because we should stay on concentration

Master-Matt:
I tested this and It works much better. Not sure why we even pressed BH again in the first place since it's on left click.


aeon0:
Since it is on left skill it will not cancel concentration. I see no harm in doing it.

Master-Matt:
Ah, I'm dumb. put BH on right lol because I never switch my left. my bad.

Issue #112, opened 2021-11-25
Template not matching for full rejuvs?


mgleed:
**Botty Version**
Master branch commit `11b55a1` (post 4.0). 1920x1080. Belt rows =4. HP=2, RV=2.

**Describe the bug**
Bot won't pick up full_rejuv potions when plain rejuv potion is disabled in params.ini and full_rejuv is enabled. Bot starts picking up both when plain rejuv is re-enabled (log file reads match for regular rejuv).

**To Reproduce**
Make sure you need rejuv potions, then watch bot during pickit when plain rejuv potion is disabled in params.ini. Might only happen in 1920x1080, haven't tested 720p

**Screenshots / Video**
```
;potions
misc_rejuvenation_potion=0
misc_full_rejuvenation_potion=1
misc_super_healing_potion=1
misc_super_mana_potion=1
```
**video**
https://user-images.githubusercontent.com/9866239/143371894-086cbef6-6f28-42d3-b1bc-8c9418573484.mp4

![Screenshot 2021-11-24 213251](https://user-images.githubusercontent.com/9866239/143371927-74721175-10da-4026-863c-06ce88b4ebcb.png)
```
;potions
misc_rejuvenation_potion=1
misc_full_rejuvenation_potion=1
misc_super_healing_potion=1
misc_super_mana_potion=1
```

**video**
https://user-images.githubusercontent.com/9866239/143371975-6a4f0076-2fd0-4830-abe6-1ca85a4b7ee8.mp4

![misc_rejuv](https://user-images.githubusercontent.com/9866239/143371993-54d27d7b-ecf4-4db3-ac7c-4159fe758ac4.png)

Graphics debugger:
![debug](https://user-images.githubusercontent.com/9866239/143373009-25562699-ecfe-407e-903e-72b9cede0522.png)

I looked at the 1920x1080 assets and they seem fine. Regular rejuv and full rejuv are equivalent pixels/in so should be appropriate for 1920.

aeon0:
Without looking into it think see the problem, and we might have that in all situations where there is a item that is a "subset" of another. e.g. 
What happens is, both are matched, but normal rejuv is slightly higher, so bitty decides it is this item and discards the other one.
We might have similar issues with superior stuff :/

mgleed:
Maybe loading template files in order from longest to shortest filename. Then maybe a way to check if a template name is a substring of an already loaded template. Tbh not super familiar with that part of src

mgleed:
[...]

PR #113, opened 2021-11-25
items overhaul for 1280_720, credits: Legit#4080

ggodev:
more tests might be needed


params.ini
```diff
@@ -80,9 +80,12 @@ pathing_delay_factor=5
...
 
 [items]
-;=====MISC=====
+;==================================================================;
+;<<<<<<<<<<<<<<<<<<<<<<<<<<<< MISC >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>;
+;==================================================================;
+
 ;potions
-misc_rejuvenation_potion=1
+misc_rejuvenation_potion=0
```
aeon0, line 88: This should be set to 1 by default. Because of this issue #112

params.ini
```diff
@@ -114,141 +117,319 @@ misc_flawless_topaz=0
...
+
+;shields
+magic_monarch=1
+
+;elite paladin shields
+magic_sacred_targe=0
+magic_sacred_rondache=0
+magic_kurast_shield=0
+magic_zakarum_shield=0
+magic_vortex_shield=0
```
aeon0, line 147: Any case a magic pala shield is worth anything?

params.ini
```diff
@@ -114,141 +117,319 @@ misc_flawless_topaz=0
...
+magic_ghost_wand=0
+magic_lich_wand=0
+magic_unearthed_wand=0
+
+;assassin katars
+magic_greater_claws=0
+magic_greater_talons=0
+magic_feral_claws=0
+magic_runic_talons=0
+magic_suwayyah=0
```
aeon0, line 172: I think shopping blue claws is a lot more effective if you looking for these.

params.ini
```diff
@@ -114,141 +117,319 @@ misc_flawless_topaz=0
...
+magic_wire_fleece=0
+magic_scarab_husk=0
+magic_archon_plate=0
+
+;gloves
+magic_bramble_mitts=0
+magic_vampirebone_gloves=0
+magic_vambraces=0
+magic_crusader_gauntlets=0
+magic_ogre_gauntlets=0
```
aeon0, line 192: As with claws, we have a shop bot for these.

params.ini
```diff
@@ -283,82 +464,396 @@ rune_31_jah=2
...
+
+;superior white
+white_superior_archon_plate=1
+white_superior_dusk_shroud=1
+white_superior_light_plate=0
+white_superior_mage_plate=1
+white_superior_great_hauberk=1
+white_superior_wire_fleece=1
+white_superior_scarab_husk=1
+white_superior_wyrmhide=1
```
aeon0, line 494: Please put all of these default to 0

aeon0 (changes requested):
Man, so many items : D I dont need all that pickit list, but it seems ppl are asking for it. So thanks to you and Legit! 

Added some comments, will test runtime. If it is still acceptable, we can merge.

params.ini
ggodev: jewelers for palas

aeon0 (approved):

PR #125, opened 2021-11-26
support list input for template_finder.search()

mgleed:


mgleed:
Merged and updated a line or two to catch up to recent updates. Tested a few runs on it tonight, going fine. Should be ready for review.

src/bot.py
```diff
@@ -267,10 +268,10 @@ def doit(self, bot: Bot):
                     return
                 bot._curr_location = Location.NIHLATHAK_PORTAL
                 wait(0.2, 0.4)
-                self.success &= bot._char.select_by_template(["A5_RED_PORTAL", "A5_RED_PORTAL_TEXT"], expect_loading_screen=True)
+                portal_found = bot._char.select_by_template(["A5_RED_PORTAL", "A5_RED_PORTAL_TEXT"], expect_loading_screen=True)
                 time.sleep(0.5)
-                self.success &= bot._template_finder.search_and_wait(["PINDLE_0", "PINDLE_1"], threshold=0.65, time_out=20)[0]
-                if not self.success:
+                pindle_start_found = bot._template_finder.search_and_wait(["PINDLE_0", "PINDLE_1"], threshold=0.65, time_out=20)
+                if not (portal_found and pindle_start_found):
```
aeon0, line 274: why you dont use self.success here anymore? It is checked afterwards to determine if we want to end run or end game.

aeon0 (changes requested):

src/ui_manager.py
```diff
@@ -144,16 +145,16 @@ def start_game(self) -> bool:
...
-            found_btn, btn_pos = self._template_finder.search("PLAY_BTN", img, roi=self._config.ui_roi["go_btn"], threshold=0.8)
+            found_btn = self._template_finder.search("PLAY_BTN", img, roi=self._config.ui_roi["go_btn"], threshold=0.8)
             score_enabled = self._template_finder.last_score
             # same as above, just with different template
             self._template_finder.search("PLAY_BTN_GRAY", img, roi=self._config.ui_roi["go_btn"], threshold=0.8)
             score_disabled = self._template_finder.last_score
             # found_btn is a bool (True or False), if a play btn was found and it is not grayed out, then we can proceed
-            found_btn = found_btn and score_enabled > score_disabled
-            if found_btn:
+            play_active = found_btn and score_enabled > score_disabled
```
aeon0, line 154: if for whatever reason play_btn_gray is not found the score will be the same. Can you reset the "last_score" in template finder search function()? Just add self.last_score = -1 whenever we call search()

aeon0 (changes requested):

src/bot.py
mgleed: I didn't see it get reused at the end, must have missed it. I only did it here because you can't do arithmetic &= on objects. Will rework according to our discord discussion.

src/ui_manager.py
mgleed: I don't think that has to do with anything I've modified here, right? If I remember right I thought I just consolidated a couple lines. score_disabled will be the score of the last attempt even if not found, if its not found it'll just show the horrible score. Seems to behave fine, it waits appropriately for the play button to activate.

could probably rework this whole section though using multiSearch and passing ["PLAY_BTN_GRAY","PLAY_BTN"] with best_match=True if desired.

aeon0: No, nothing you introduced, just thought that would be nice  fix while we are at it : )

src/bot.py
mgleed: Now that I'm using .valid I can use the original logic.

src/ui_manager.py
mgleed: I ended up reworking this, have only tested one time. Hope it works :p

aeon0 (approved):

PR #132, opened 2021-11-27
Make it possible to initialize Screen ingame

reqyl:
This fixes broken shopping script due to only lobby being supported right now. This also fixes standalone scipts node recorder. Both resolutions are supported


src/screen.py
```diff
@@ -26,14 +26,28 @@ def __init__(self, monitor: int = 0):
...
+            _, max_val_ingame, _, max_pos_ingame = cv2.minMaxLoc(res_ingame)
+
+            # We are in game
+            if max_val_ingame > max_val:
+                max_val = max_val_ingame
+                offset_x, offset_y = max_pos_ingame
+                if self._config.general['res'] == "1280_720":
+                    max_pos = (offset_x - 625, offset_y - 670)
+                else:
+                    max_pos = (offset_x - 938, offset_y - 1004)
```
aeon0, line 48: Can you add these coordinates as params to the game.ini? Then you don't need an if here and can just use:
max_pos = (offset_x - config.ui_pos["ingame_ref_x"], offset_y - config.ui_pos["ingame_ref_y"])

aeon0 (changes requested):
Great idea, great PR. Little change to parameterize hardcoded values and good to go.

src/screen.py
reqyl: Ah good idea. Nice I see those coordinates are auto-scaled so we only need one tuple.

aeon0 (approved):

PR #134, opened 2021-11-28
autodownload nuitka dependencies for compile

mgleed:


aeon0:
What kind of errors did that solve? Just so I can tell them "fixed on the latest master!" : )

mgleed:
> What kind of errors did that solve? Just so I can tell them "fixed on the latest master!" 

No errors per-se, but you have to hit "yes" to probably 5 or so dependencies the first time you compile from source, took me like 30 minutes before I got it all done because I kept expecting it to go to completion.

PR #145, opened 2021-11-29
Update potion needs after drinking which allows accurate potion pickups after fights.

jrzheng:


jrzheng:
turns out that this is already getting addressed, but I think it should be a min and not a max.

aeon0:
Good catch. Missed that on my review, thx.

PR #146, opened 2021-11-29
blacklisted superior flameberge, as it gets recognized for superior flail

ggodev:


aeon0:
Yep, also rare battle bow with rare battle boots I think...

ggodev:
haven't seen those, ill keep an eye

GeminiCCCC:
> 

rare battle bow +1

PR #147, opened 2021-11-29
Gray button handling

haver007:
Added gray button handling. Instead of going to an error. The bot will now retry to join a game after 5 seconds.


aeon0:
Its already handled in L151

PR #148, opened 2021-11-29
Full Standalone Shopper Script for Anya IAS Gloves and Claws

SwePow:


src/template_finder.py
```diff
@@ -177,6 +177,7 @@ def search(
 
         if img.shape[0] > template.shape[0] and img.shape[1] > template.shape[1]:
             res = cv2.matchTemplate(img, template, cv2.TM_CCOEFF_NORMED)
+            self.last_res = cv2.matchTemplate(img, template, cv2.TM_CCOEFF_NORMED)
```
aeon0, line 180: Why we call this twice? That's not good for runtime : ) Just do self.last_res = cv2.matchTemplate() and use self.last_res instead of res.

src/shopper.py
```diff
@@ -0,0 +1,318 @@
...
+    3.) Stand close to Anya and the red portal.
+    4.) While being ingame, press resume_key (default F11) to start the shopping, and exit_key (default F12) to stop it.
+    """
+
+    def __init__(self, config):
+        
+        # Configurable part, dont touch anything above
+        # Set look_for variables to False if you dont like your personal shopper to look for these
+        # Obviously something need to be set to True, or your shopper will be very confused
+        # For the claw scores use 7 if you are happy with any + to traps or +2 assassin, 9 if you want at least +2 assassin or two useful trap stats, 11 if you
```
aeon0, line 76: 11 if you... what a clif hanger! :D

src/shopper.py
```diff
@@ -0,0 +1,318 @@
...
+                        normalize_monitor=True,
+                    )
+                    if gg_gloves_found:
+                        mouse.click(button="right")
+                        send_discord(f"Bought awesome IAS/+3 gloves!", config.general["custom_discord_hook"])
+                        Logger.info("IAS/+3 gloves bought!")
+                        self.gloves_bought += 1
+                        time.sleep(1)
+
+# Enable the following, if you are fine with +2 gloves
```
aeon0, line 153: indent

src/shopper.py
```diff
@@ -0,0 +1,318 @@
...
+                            normalize_monitor=True,
+                        )
+                        if g_gloves_found:
+                            mouse.click(button="right")
+                            send_discord(f"Bought some decent IAS/+2 gloves", config.general["custom_discord_hook"])
+                            Logger.info("IAS/+2 gloves bought!")
+                            self.gloves_bought += 1
+                            time.sleep(1)
+
+# Select Weapons section
```
aeon0, line 171: indent

aeon0 (changes requested):

aeon0:
Also, can we move the "shopper.py" in the shop folder?

src/shopper.py
```diff
@@ -0,0 +1,318 @@
...
+        self.look_for_plus_2_gloves = True
+        self.look_for_plus_3_gloves = True  
+        self.look_for_trap_claws = True
+        self.trap_claw_min_score = 11
+        self.look_for_melee_claws = True
+        self.melee_claw_min_score = 11
+        
+        # Dont touch anything below here
+        
+        self._screen = screen.Screen(config.general["monitor"])
```
aeon0, line 87: Pass the screen instance to the constructor instead of creating a new one. Otherwise two screen instances are created.

aeon0 (changes requested):

src/shopper.py
SwePow: Oh, I dont know Python enough for that - I am a pseudo coder. Can someone help?

aeon0: I can do : )

src/shopper.py
SwePow: Fixed and improved!

src/shopper.py
SwePow: Removed, the look_for_plus_2_gloves kinda makes the comment redundant.

src/shopper.py
SwePow: Fixed!

src/template_finder.py
SwePow: Right, that is better. I decided to keep both so I dont break anyone elses use. Fixed according to your suggestion.

SwePow: Would self.last_res = res work even?

aeon0 (approved):

PR #151, opened 2021-11-30
Revert incorrect pot increment

reqyl:
I think this was correct before last commit.

### DIFF
--- a/src/belt_manager.py
+++ b/src/belt_manager.py
@@ -82,7 +82,7 @@ class BeltManager:
                 else:
                     Logger.debug(f"Drink {potion_type} potion in slot {i+1}. HP: {(stats[0]*100):.1f}%, Mana: {(stats[1]*100):.1f}%")
                     keyboard.send(self._config.char[key])
-                self._pot_needs[potion_type] = min(self._config.char["belt_rows"], self._pot_needs[potion_type] + 1)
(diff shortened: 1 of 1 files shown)


aeon0:
I dont get it. I think your alst change was good, no need to revert.
When you keep adding you don't ever want to pick up more pots than you have rows in your belt. 

Ah now that I write it. It all fails when you select to have multiple columns used for one type :man_facepalming:
IMHO the min is not needed, but doesn't hurt either.

jrzheng:
I don't get this neither. BTW I was the one who made the fix. It's incorrect to use max, you'd end up picking up more pots than belt rows.

jrzheng:
> Ah now that I write it. It all fails when you select to have multiple columns used for one type 🤦‍♂️ IMHO the min is not needed, but doesn't hurt either.

Good point about multiple columns. I believe the fix should be 

self._pot_needs[potion_type] = min(self._config.char["belt_rows"] * rows_left[potion_type], self._pot_needs[potion_type] + 1)

define rows_left earlier as

        rows_left = {
            "rejuv": self._config.char["belt_rejuv_columns"],
            "health": self._config.char["belt_hp_columns"],
            "mana": self._config.char["belt_mp_columns"],
        }

PR #155, opened 2021-11-30
fix(sorc): hold mouse during attack delay

dcolondev:
Small change to keep the mouse held down during this attack delay, seems to speed up the attack sequence by a few frames and I haven't had any issues in my testing.


jagarop:
The only problem would be hit recovery. Maybe adopt this solution and reduce time of each cast, increasing atk_len_pindle default?

aeon0:
I dont see that issue. Because while previously we would just click we now hold. So the same frequency of casting the skill as previously.

PR #159, opened 2021-12-01
Out of gold fix

Master-Matt:
Added ability for ItemFinder UIManager and PickIt to use passed in Config and allowed config to be init'd with runtime changes....
All to fix out of gold/unable to repair loop/bug. Overkill? Yes. Completely unnecessary ? Probably. Useful? Maybe...


src/bot.py
```diff
@@ -33,8 +33,8 @@ def __init__(self, screen: Screen, pick_corpose_on_start: bool = False):
         self._game_stats = GameStats()
         self._game_recovery = GameRecovery(self._screen)
         self._template_finder = TemplateFinder(self._screen)
-        self._item_finder = ItemFinder()
-        self._ui_manager = UiManager(self._screen, self._template_finder)
+        self._item_finder = ItemFinder(self._config)
+        self._ui_manager = UiManager(self._screen, self._template_finder,self._config)
```
aeon0, line 37:
```suggestion
        self._ui_manager = UiManager(self._screen, self._template_finder, self._config)
```
there are a few more cases of missing whitespace

src/bot.py
```diff
@@ -198,6 +202,22 @@ def on_maintenance(self):
...
+                self._ui_manager.close_vendor_screen()
+                config_overrides = {"misc_gold":"1","stash_gold":"1"}
+                self._config = Config(False,config_overrides)
+                self._item_finder = ItemFinder(self._config)
+                self._ui_manager = UiManager(self._screen, self._template_finder,self._config)
+                self._pickit = PickIt(self._screen, self._item_finder, self._ui_manager, self._belt_manager, self._game_stats)
+                self.out_of_gold = True
+                Logger.warning("Out of gold. Adding gold to pickit until next bot restart.")
+                self._do_runs["run_pindle"] = False
+                Logger.info("Disabling Pindle to get more gold per run")
```
aeon0, line 216: I wouldnt do that. What if user only runs pindle? Also this will be overwritte in game_end() by whatever is in self._routes_config.

src/bot.py
```diff
@@ -67,6 +67,9 @@ def __init__(self, screen: Screen, pick_corpose_on_start: bool = False):
         self._pausing = False
         self._current_threads = []
         self._no_stash_counter = 0
+        self.out_of_gold = False
+        self.wait_to_repair = 0
+
```
aeon0, line 72:
```suggestion
        self.wait_to_repair = 0
```
Remove extra new line

src/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.4.2-dev'
\ No newline at end of file
+__version__ = '0.4.3-dev'
```
aeon0, line 1: dont change version

src/bot.py
```diff
@@ -198,6 +202,22 @@ def on_maintenance(self):
...
                 self._ui_manager.close_vendor_screen()
                 self._tps_left = 20
+            elif not self.out_of_gold: #You are out of gold. Adding gold to pickit until you can repair
+                wait(0.1, 0.2)
+                self._ui_manager.close_vendor_screen()
+                config_overrides = {"misc_gold":"1","stash_gold":"1"}
+                self._config = Config(False,config_overrides)
+                self._item_finder = ItemFinder(self._config)
+                self._ui_manager = UiManager(self._screen, self._template_finder,self._config)
+                self._pickit = PickIt(self._screen, self._item_finder, self._ui_manager, self._belt_manager, self._game_stats)
```
aeon0, line 212: instead of creating new instances you could also write a setter for the config. That way if the constructor changes to any of these classes you dont have to adapt these lines.

src/pickit.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, screen: Screen, item_finder: ItemFinder, ui_manager: UiManage
         self._belt_manager = belt_manager
         self._ui_manager = ui_manager
         self._game_stats = game_stats
-        self._config = Config()
+        self._config = self._item_finder._config
```
aeon0, line 23: Underscore means that it is supposed to be a private variable. Either remove the underscore of config in item finder or write a getter for it.

src/bot.py
```diff
@@ -198,6 +202,22 @@ def on_maintenance(self):
...
+                self.out_of_gold = True
```
aeon0, line 213: I dont see where we would ever set this back to False

aeon0 (changes requested):
So overall the change is cool. I am just not sure I want to add that complexity for this edge case... If you run out of gold to buy tp tombs you most probably have bigger issues with your build :D

Also there is already at tomb template in the asset folder, it is just speed wrong.

src/version.py
Master-Matt: Will build with current version in the future

src/bot.py
Master-Matt: Yeah, I changed this to randomize runs config override. Because if just running pindle, hdin has more gold repair cost per run than just pindle can provide.

src/bot.py
Master-Matt: The comment was not changed. The warning says that it will change upon bot restart after destroying and creating a new bot instance.

src/pickit.py
Master-Matt: Yeah, I just added a separate config to pass in for this

src/bot.py
Master-Matt: This I don't quite understand. When you init ItemFinder it grabs the config, so wouldn't you always need to re-init, or would that mean write a private config setter for each class that does the equivalent of re-initing?

Master-Matt:
Overall, one of the things I was wondering about is why almost all classes create their own instances of config.
I guess part of my idea is that the other classes can share a config so if, in the future, we want to add config logic that alters the config based on some parameter we set, we just only have to change one object.

PR #162, opened 2021-12-01
Add support for forcing the bot to re-apply buffs prior to running sh…

dennislbc:
Players can now choose to force the bot to re-apply their buffs after taking the frigid WP to ensure their buffs are still up when they run eld / shenk.


aeon0 (changes requested):
When randomizing runs it might be that you run of cta for pindle. Or would you say you don't care in this case? If not a "force_rebuff" without the shenk specification would be nicer I guess.

After discussion we said we want to have it both ways. Then I would change the name to "force_pre_buff" and it would force a prebuff after every run.

aeon0:
Has  been added in a different PR

PR #167, opened 2021-12-01
Add in logic when botty is paused and not trigger max game length

egut125:


src/game_stats.py
```diff
@@ -57,10 +59,34 @@ def log_end_game(self):
     def log_failed_run(self):
         self._runs_failed += 1
 
-    def get_current_game_length(self):
+    def pause_timer(self):
+        """ Pauses the timer """
```
aeon0, line 63: No need to add comments if they are the same as the function name. There is not going to be any quality team that forces us to have every function documented with comments :yum:

src/game_stats.py
```diff
@@ -57,10 +59,34 @@ def log_end_game(self):
...
+        if self._timer is None:
+            return
+        if self._paused:
+            return
```
aeon0, line 67: You can combine these two `if self._timer is None or self._paused:`

src/game_stats.py
```diff
@@ -57,10 +59,34 @@ def log_end_game(self):
...
+        Logger.info(f'Pausing timer')
```
aeon0, line 68: Make this a debug message maybe. There is already a info message in bot.py when it gets paused

src/game_stats.py
```diff
@@ -57,10 +59,34 @@ def log_end_game(self):
...
+        self._timepaused = time.time()
+        self._paused = True
+
+    def resume_timer(self):
+        """ Resumes the timer by adding the pause time to the start time """
+        if self._timer is None:
+            return
+        if not self._paused:
+            return
```
aeon0, line 77: As above, can be combined

aeon0 (changes requested):
Few small change requests. But logic looks good!

aeon0 (approved):

aeon0:
Thx!

PR #170, opened 2021-12-02
Template alpha channel

mgleed:
Add option to use a template's alpha channel as a mask for a template match. Anywhere in an image that has complete transparency (channel 4 = 0) will have that region ignored during template matching using a mask for `cv2.matchTemplate()`. Just need to pass the parameter `alpha = True` to `load_template()`.

One example was included here that adds the ability for player to use any type of merc by making the inner portrait of the merc template transparent.

If desired, this could be adapted to be the default behavior rather than having to pass `alpha=True`, but then any instance where `template_match` is used (outside of `template_finder.search()`) would need to be adapted because `cv2.matchTemplate()` doesn't accept RGBA format. I made it parameter-based more out of convenience.


src/template_finder.py
```diff
@@ -190,14 +190,20 @@ def search(
             template_match = TemplateMatch()
             scale = scales[count]
 
+            mask = None
```
aeon0, line 193: Can you add one short comment above this code what it does. something like `# Use alpha channel to create a mask for alpha == 0. Since it does something different than the name suggests in the first place. So whenever code is not "self documented" e.g. by variable names, it makes actually sense to add comments.

aeon0 (changes requested):
The concept of using alpha channel as a mask is pretty nice! I dont yet get the specific usecase of the merc. You removed the merc template? Shouldnt you add a new template?

mgleed:
This is a case of git doing git things, dunno why it didn't push the replacement merc.png. Will fix that and then write comment. Thanks!

mgleed:
@aeon0 It says this PR has deleted merc.png, but if you browse the src of the branch it's definitely there ... idk what else to do. The comment is there now.

mgleed:
Latest commit I cleaned up all the NPC templates--hopefully this means they might detect faster? I tested repair, revive merc, and purchase new TP's at it worked fine. Didn't test anya.

Also I put transparency on the you_have_died image so now it should be easily detected using `template_finder.search()` instead of doing color operations. Code adjusted accordingly (feels so much better to delete lines than it does to add lines!) I tested it once at it worked nicely.

aeon0 (approved):

PR #174, opened 2021-12-02
Revert npc multi tempalte search

aeon0:
Reverts parts of #171 
Sry for the bad review @mgleed. Apart from the issue with the passed argument, I think we have to keep it split because we do gold and white filtering separately.


mgleed:
Oh yeah, thanks. I had tested and realized that wasn't correct and reverted the same change as you but of course forgot to hit save in vscode. Thanks,

PR #182, opened 2021-12-02
skip gold stashing when there's no gold in inventory

navy91:


src/ui_manager.py
```diff
@@ -282,14 +282,19 @@ def stash_all_items(self, num_loot_columns: int, item_finder: ItemFinder):
...
         if self._config.char["stash_gold"]:
-            x, y = self._screen.convert_screen_to_monitor(gold_btn.position)
-            mouse.move(x, y, randomize=4)
-            wait(0.1, 0.15)
-            mouse.press(button="left")
-            wait(0.25, 0.35)
-            mouse.release(button="left")
-            wait(0.4, 0.6)
-            keyboard.send("enter") #if stash already full of gold just nothing happens -> gold stays on char -> no popup window
+            inventory_no_gold = self._template_finder.search_and_wait("INVENTORY_NO_GOLD", roi=self._config.ui_roi["gold_btn"], time_out=1)
```
aeon0, line 285: I am right now not a 100% sure if the gold_btn roi is large enough to include the gold amount . You can double check it by running screen.py which will visualize the regions.

I would just call search() instead of search and wait as we do not expect any changes in the scene. Will speed things up a bit.

aeon0 (changes requested):

aeon0 (approved):

PR #184, opened 2021-12-03
Quick Cast (WIP)

jagarop:
didn't made changes in hammerdin.py because i don't know how it will perform with auras.

but the idea is here

more competent coders like @aeon0 will implement this in seconds 👯


franklin050187:
maybe should be an option, if quickbar is set to 1 then use hotkey and no mouse, otherwise run as currently

aeon0:
> maybe should be an option, if quickbar is set to 1 then use hotkey and no mouse, otherwise run as currently

There will be for sure no support for both. If we switch we remove the other logic. That would be too much maintenance and code switching. Plan is to only use quick cast at some point.

PR #187, opened 2021-12-03
Tome changes

cutebeast:
no longer need to have 20 tp when start as now it will check for tome condition especially when tome = 0


src/ui_manager.py
```diff
@@ -1,442 +1,456 @@
...
+        if not tp_tome.valid:
+            return False
+        return True
+
+    # Start Section by Cutebeast: Checking for Tome Condition
+    # Return True if no more TP
+    def no_more_tp(self) -> bool:
+        keyboard.send(self._config.char["tp"])
+        wait(0.02, 0.05)
+        tome_active = self._template_finder.search_and_wait("TOME_ACTIVE", roi=self._config.ui_roi["skill_right"], time_out=4)
```
aeon0, line 436: So we check if we have an inactive tomb, right? Since the template is red. That means we have an added delay of 4 seconds after each run?

Also having a "no" in the function and returning "False". Usually it is easier to read to have "has_tps()" and return True.

aeon0 (changes requested):
Something seems to be off with line endings. The whole code file shows up as changed

src/ui_manager.py
cutebeast: ok i will fix it up tomorrow

cutebeast:
Updating new file

PR #200, opened 2021-12-04
Tome changes & IP Dclone changes

cutebeast:
For Tome - user don't need to start with Full tome anymore
For Dclone IP - user need to key in the reqion IP to target & the hot IP, once current game ip and hot ip matches the bot will wait for user to key in Enter to stop the bot


cutebeast:
Will fork new for PR

PR #205, opened 2021-12-04
No item pickup

Master-Matt:
Implemented simple /nopickup check.


src/ui_manager.py
```diff
@@ -448,6 +448,29 @@ def has_tps(self) -> bool:
         else:
             return False
 
+    def enable_no_pickup(self) -> bool:
+        """
+        Checks the best match between enabled and disabled an retrys if already set.
+        :return: Returns True if we succesfully set the nopickup option
+        """
+        keyboard.send('enter', do_release=True)
```
aeon0, line 456: keyboard.send() has do_press and do_release at True per default

aeon0 (changes requested):
Good stuff. One thing. In your screenshots you have a black background on the messages. This background can be toggeled on and off in the d2r gameplay settings. Currently botty does not adjust this, so depending on the user settings the template matching might have problems.

Can you add this setting to the assets/d2r_settings.json?

aeon0 (approved):

PR #211, opened 2021-12-05
Force pre buff

dennislbc:
Player can choose force pre buff in params to ensure the bot uses buffs/cta prior to all runs. 

Previously, if a player's CTA expired too quickly they would be doing Shenk without CTA active. Force buff makes the bot run the CTA/Buff routine upon entering pindle run and after taking the frigid WP.


dennislbc:
Fix for my last broken PR

PR #214, opened 2021-12-06
add support for custom offsets

mgleed:
for those struggling to get auto offset detection to work

### DIFF
--- a/params.ini
+++ b/params.ini
@@ -78,6 +78,8 @@ redemption=
 ;===== don't touch unless you know what you're doing =====
 pathing_delay_factor=5
 template_threshold=0.68
+offset_left=
+offset_top=
 
 [items]
 ;==================================================================;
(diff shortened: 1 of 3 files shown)


aeon0:
I dont know. If someone struggels to get auto offset work, isn't that a hint there is something wrong that will not be fixed by adding offsets?

mgleed:
> I dont know. If someone struggels to get auto offset work, isn't that a hint there is something wrong that will not be fixed by adding offsets?

Your point is well taken, do what you will. Had a couple saying they're stuck in 0.3 because botty 0.4 won't get past the offset detect log and they showed me screenshots. In my laziness I figured this fix would be easier than having to do video chat or something for everyone with the issue.

aeon0:
I added some debug messages about the maximum score if the templates were not found and lowered it a bit. Also added a warning if your scores are lower than usual. That should give a decent hint at what the problem might be.

I am sticking to my opinion that fixing the symptom will not fix the issue for them. You will just here follow up issues like. Stuck at eld waypoint etc.

mgleed:
I can see where you're coming from. I don't know if I would have tried this hadn't we done something similar just this week, however. Would you agree with this same reasoning for #206 ? Either there's inconsistency with template matching (maybe platform based?) or the users aren't setting their configurations right, but either way a symptom was addressed at the cost of a feature.

PR #225, opened 2021-12-07
Fix item pickup behind obstacles

aeon0:
Quite a bit hacky... timer will check if we try to pick up the same item over and over again. After 4 sec it will try to force a move (with teleport this can fix the issue), after another 4 sec it will skip this item type for this round.


GeminiCCCC:
nice fix

PR #231, opened 2021-12-07
fix game start in act 4 (adj start tempates)

bobcho:


aeon0:
thx, pathing issue we can fix afterwards

PR #250, opened 2021-12-10
Separate gold and item stashing

Master-Matt:
Implemented gold and item stash counters. botty will quit either if stash space for items is gone or all gold stash and character is full.
For character we just check 0000 so it could theoretically quit early if all tabs are full and we hit any 10k multiple past 100k. but it's unlikely and still we get almost all the gold we want.


src/ui_manager.py
```diff
@@ -303,7 +303,32 @@ def stash_all_items(self, num_loot_columns: int, item_finder: ItemFinder):
                 mouse.release(button="left")
                 wait(0.4, 0.6)
                 keyboard.send("enter") #if stash already full of gold just nothing happens -> gold stays on char -> no popup window
+                wait(1.0, 1.2)
+                inventory_no_gold = self._template_finder.search("INVENTORY_NO_GOLD", self._screen.grab(), roi=self._config.ui_roi["inventory_gold"], threshold=0.97)
+                if not inventory_no_gold.valid:
+                    Logger.info("Stash tab is full of gold, selecting next stash tab.")
+                    self._curr_stash["gold"] += 1
+                    if self._curr_stash["gold"] > 3:
+                        inventory_full_gold = self._template_finder.search("INVENTORY_FULL_GOLD", self._screen.grab(), roi=self._config.ui_roi["inventory_gold"], threshold=0.98)
```
aeon0, line 312: Wouldnt this also trigger if by chance char has exactly 10000 gold? But I understand that's the best we can do cause we don't know char lvl and that would be a edge case that might not be worth the hustle.

aeon0:
LGTM. Ui manager and specifically the stashing function is getting messy... (not your fault). Will try to split things up a bit in the future. Thx for the implementation, was probably a pain to test...

PR #258, opened 2021-12-11
Use message templates to support other hooks

nekolr:


params.ini
```diff
@@ -11,8 +11,11 @@ graphic_debugger_key=f10
 logg_lvl=debug
 randomize_runs=0
 difficulty=hell
-custom_discord_hook=
+custom_message_hook=
 discord_status_count=20
+message_headers=
+message_body_template={{"content": "{msg}"}}
+message_highlight=1
```
aeon0, line 18: Maybe these would be good at the [advanced_options] section. Most users already struggle creating a discord hook :D

src/message_sender.py
```diff
@@ -0,0 +1,44 @@
+from config import Config
+import json
+import requests
+
+
+class MessageSender:
```
aeon0, line 6: Regarding naming. What you think about reducing naming redundancy. I am thinking something like:
`Messenger.send()`.

and yes I am well aware I also don't always adhere to this in my own code...

aeon0 (changes requested):
Thx for the PR. Looks good, minor stuff. I'd be curious what kind of API endpoint you are using other than discord.

The failed CI is not your fault... I merged sth stupid earlier, when you merge current master you should be all good, sry bout that.

nekolr:
thx for your review, i will think about what you say.

aeon0:
Why close PR? You can always push your changes to it. I like the feature : (

nekolr:
> Why close PR? You can always push your changes to it. I like the feature : (

hah, i made some changes

PR #265, opened 2021-12-11
Use message template to support other hooks

nekolr:


aeon0:
Good work! Thank you.

PR #268, opened 2021-12-12
Allow item = 3 in pickit to make ethereal required to stash

mgleed:
Simple PR, feature requested by a few. I don't love how the loop turned out so if you see a cleaner way to do it, feel free to adjust.


aeon0:
I dont like it. But I also dont see another quick and easy way to implement it. So it's ok for a workaround for now.

PR #286, opened 2021-12-14
Allow use of "shift" for stand still hotkey

mgleed:
This PR is dedicated to all of us (me) who struggle on the daily (me) to remember to hold down caps lock instead of shift after decades of muscle memory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ run_shenk=0
 | merc_chicken       | Will chicken (leave game) when merc health percentage drops below set value, range 0 to 1. Set to 0 to not chicken. |
 | show_items         | Hotkey for "show |
 | inventory_screen   | Hotkey to open up |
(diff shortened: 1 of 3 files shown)


aeon0:
The problem is if now you press standstill e.g. during throwing hammers and you get low on life and want to drink a healing pot, you will give it to the merc.
So if we wanted to allow shift for standstill, we would have to detect that during the attack sequence, stop the attack, drink pot, continue attack.

mgleed:
> The problem is if now you press standstill e.g. during throwing hammers and you get low on life and want to drink a healing pot, you will give it to the merc. So if we wanted to allow shift for standstill, we would have to detect that during the attack sequence, stop the attack, drink pot, continue attack.

Ooh, yeah, very true. Guess I was reaching for straws lol.

PR #288, opened 2021-12-14
Barbarian support (currently War Cry/singer only) with Horking

tw1st40:
Horking (with weapon swap) and some initial assistance thanks to Outburst. Works without Enigma/tele by using leap on left skill. Currently working with War Cry.


src/ui/ui_manager.py
```diff
@@ -77,6 +77,15 @@ def is_right_skill_selected(self, template_list: List[str]) -> bool:
                 return True
         return False
 
+    def is_left_skill_selected(self, template_list: List[str]) -> bool:
+        """
+        :return: Bool if skill is currently the selected skill on the right skill slot.
```
aeon0, line 82:
```suggestion
        :return: Bool if skill is currently the selected skill on the left skill slot.
```

src/char/i_char.py
```diff
@@ -172,6 +172,40 @@ def _pre_buff_cta(self):
             else:
                 Logger.warning("Turns out weapon switch just took a long time. You ever considered getting a new internet provider or to upgrade your pc?")
 
+    def _do_hork(self, hork_time: int):
```
aeon0, line 175: This is a barb specific skill and should be placed in the barbarian.py file

config/params.ini
```diff
@@ -1,6 +1,6 @@
 ; There is detailed documentation for each parameter in the README.md
 [general]
-name=Botty
+name=Botty Barbarian
```
aeon0, line 3: Please change only change the default values of already existing params if really needed.

aeon0 (changes requested):
Nice, some small remarks.

aeon0 (approved):

PR #290, opened 2021-12-15
Support Greater Potion

yhkinsar:
Allowing belt management also work with Greater health potion and Greater mana potion.
For some people who is really bad luck and not getting any super potion for around 10 games. (which is me)

Which should make botty work longer and even able to work with normal level.


aeon0:
That should do it. Thx

PR #295, opened 2021-12-15
Add require/exclude properties to pickit.ini when stashing items. Also more item templates.

mgleed:


mgleed:
i think one more feature i'd like to implement before merging would be somehow doing AND rather than just OR, but would have to rewrite some stuff. I'd really like to make it only include eth and socketed polearms, for example, but right now it just checks for either socketed or eth
maybe like item=1,and(eth, socketed_4)

mgleed:
OK, this is ready to test again. Legit is gonna review the pickit.ini given latest change to logic

mgleed:
Should be fairly steady now, please review the code and changes @aeon0 . Major changes include:
- implementing required and exclusion properties for every item to be stashed
- and/or boolean logic added for these properties
- multiple properties including ethereal, socketed, +skills, etc., see assets/item_properties for supported properties
- renamed all item assets from eth_ to gray_ to remove ambiguity (not all gray items are ethereal)
- added many rare items for those who like rare ethereals
- Legit liked being able to comment out lines in pickit.ini, please review and adjust as you see fit

will be breaking for most people's pickit configs

mgleed:
Found AND() wasn't working correctly in some cases. Seems resolved now.

config/pickit.ini
```diff
@@ -1,31 +1,55 @@
...
+; "gray_thresher=1, and(ethereal, socketed_4)" requires ethereal AND 4-sockets
+; "gray_thresher=1, ethereal, (socketed_1, socketed_2, socketed_3)"  requires ethereal;
+;    will discard any with 1-3 sockets. This will pickup non-socketed ethereal's.
+; "gray_weapon_grand_matron_bow=1, and(3_bow_skills, enhanced_damage, socketed_4)" will pickup but stash
+;    only GMB that have 3+ bow skills AND enhanced damage modifier AND 4 sockets
+
+; regarding ethereal items:
+; if you want to keep ONLY ethereal, add ethereal after the first comma not in () like `item=1, ethereal`
+; if you want to discard ethereal, add ethereal after the second comma not in () like `item=1,, ethereal`
+; if you want both ethereal and non-ethereal items, do not use the ethereal property
```
aeon0: I would move this block above examples and I don't understand what the "not in () like" is supposed to mean. Add maybe a "blueprint" example:
item_name=pickit_type, required_properties, discarded_properties

config/pickit.ini
```diff
@@ -1,31 +1,55 @@
...
 misc_rejuvenation_potion=0
 misc_full_rejuvenation_potion=1
 misc_super_healing_potion=1
 misc_super_mana_potion=1
 misc_greater_healing_potion=0
 misc_greater_mana_potion=0
 
-;other
-misc_tp_scroll=0
+;MISC
```
aeon0: whats with the upper cases? But ok, dont mind ^^

config/pickit.ini
```diff
@@ -1,31 +1,55 @@
...
+;misc_tp_scroll=1
```
aeon0: I vote for removing tp_scrolls and also its asset. With buying tps I don't see any usecase picking them up.

src/config.py
```diff
@@ -155,6 +164,22 @@ def __init__(self, print_warnings: bool = False):
             "melee_min_score": int(self._select_val("claws", "melee_min_score")),
         }
 
+    def parse_item_config_string(self, key: str = None) -> ItemProps:
+        item_props = ItemProps()
+        # split string by commas NOT contained within parentheses
+        item_string_as_list = re.split(r',\s*(?![^()]*\))', self._select_val("items", key).upper())
```
aeon0: https://i.redd.it/an2lhgcr9kc61.png

Still love it : )

src/config.py
```diff
@@ -155,6 +164,22 @@ def __init__(self, print_warnings: bool = False):
...
+        clean_string = [re.sub(r'|'.join(map(re.escape, trim_strs)), '', x).strip() for x in item_string_as_list]
+        item_props.pickit_type = int(clean_string[0])
+        try:
+            item_props.include = clean_string[1].split(',') if clean_string[1] else None
+            item_props.include_type = "AND" if "AND" in item_string_as_list[1] else "OR"
+        except: pass
+        try:
+            item_props.exclude = clean_string[2].split(',') if clean_string[2] else None
+            item_props.exclude_type = "AND" if "AND" in item_string_as_list[2] else "OR"
+        except: pass
```
aeon0: I dont see when these would fail, why the try except? It's also a bit risky to hide unwanted errors. If we are allow a specific error, e.g. ValueError, specify that one. Right now if there is an issue with these statements we will never see it.

aeon0: Maybe even add print the error at least to the debug log

src/ui/ui_manager.py
```diff
@@ -271,10 +271,67 @@ def _keep_item(self, item_finder: ItemFinder, img: np.ndarray) -> bool:
...
+        for x in original_list:
+            if "potion" in x.name: continue
+            include_props = self._config.items[x.name].include
+            exclude_props = self._config.items[x.name].exclude
+            if not (include_props or exclude_props):
+                Logger.debug(f"{x.name}: Stashing")
+                filtered_list.append(x)
+                continue
+            include = True
+            include_bool_type = self._config.items[x.name].include_type
```
aeon0: I find the name confusing. `include_bool_type` suggests that this variable is a bool while it really is a str ["OR", "AND"]

src/config.py
mgleed: isn't this the purpose of stack overflow? 

Also reminds me of this in my field of work: [:)](https://scontent-atl3-1.xx.fbcdn.net/v/t1.6435-9/79532782_1867938783341570_1908639344143368192_n.jpg?_nc_cat=100&ccb=1-5&_nc_sid=8bfeb9&_nc_ohc=1rJX9JU7mJsAX89TOhS&_nc_ht=scontent-atl3-1.xx&oh=00_AT_knVpqruEOvw3aN1qA0gVKiEcKE2nTdM81YupqdREyuw&oe=61E62CAB)

src/config.py
mgleed: this is to handle pickit.ini lines that have variable # of fields.
so I can say `=1` `=1, ethereal` and `=1, socketed, ethereal`

a little lazy way to go about it I guess =/

aeon0: Okay, then do except IndexError. This way you will still see errors that are not expected.

aeon0:
Thx for enduring my nitpicking.

PR #299, opened 2021-12-17
adding more uniq items

yhkinsar:
Just added more Exceptional & Elite uniq items but not include any change on weapon.


aeon0:
Sure, why not. But you forgot Stoutnail...

yhkinsar:
Why I miss that!? nvm I will add it when I work on weapon list.

Best regards,
John Yue

PR #303, opened 2021-12-19
Correctly Total Failed Runs, Only Set location to shenk when battling…

egut125:
…, not on run start


src/game_stats.py
```diff
@@ -148,6 +148,7 @@ def _create_msg(self):
             totals["chickens"] += stats["chickens"]
             totals["deaths"] += stats["deaths"]
             totals["merc_deaths"] += stats["merc_deaths"]
+            totals["failed_runs"] += stats["failed_runs"]
```
aeon0, line 151: Hmm, that was an easy solution to keep the number of failed runs to a minimum... oh well.

PR #304, opened 2021-12-19
Nihlathak run changed to use the correct path if the template 'NI2_A_SAFE_DIST_2' was detected

stephan-01010011:
- Template assets renamed to use an uniform suffix. _save_dist, _noattack
- EyeCheckData class enhanced to contain a destination static path definition that is used instead of the template name


aeon0:
Right. Good catch.

PR #310, opened 2021-12-20
Config discord condensed

egut125:
…d status report

Defaults to 0


aeon0:
why not : )

egut125:
> why not : )

Haha got too many questions in the discord about what the letters meant and if we could go with more words.... so I gave in but still left it so I could have the small version.

PR #311, opened 2021-12-21
tome bug workaround

kozfelipe:
tome bug workaround


aeon0:
that should do for now, thx

PR #322, opened 2021-12-24
Arcane Sanctuary demo for tweaks

kozfelipe:
plus working act 2 town manager


aeon0:
Cool. Just browsing over it I am having a good first impression of the implementation : ) Will test more once I get the time!

src/char/sorceress/sorceress.py
```diff
@@ -32,6 +32,40 @@ def pick_up_item(self, pos: Tuple[float, float], item_name: str = None, prev_cas
             return cast_start
         else:
             return super().pick_up_item(pos, item_name, prev_cast_start)
+
```
kozfelipe, line 35: support for telekinesis

jrzheng:
This works well! I cherry picked this change and added the ability to TP back to town and take WP so we can check all 4 directions. But I'm not sure if it's better to use TP or backtrack in arcane?

Please note that the summoner drops Key of Hate. The template here is for Key of Terror.

src/health_manager.py
```diff
@@ -48,7 +48,7 @@ def reset_chicken_flag(self):
 
     def update_location(self, loc: Location):
         if loc is not None and type(loc) == str:
-            bosses = ["shenk", "eldritch", "pindle", "nihlatak", "trav"]
+            bosses = ["shenk", "eldritch", "pindle", "nihlatak", "trav", "arcane"]
```
jrzheng, line 51: Looks like this should be "arc" instead to match locations such as "a2_arc_start". I was wondering why the potions are not working :)

kozfelipe: thanks

kozfelipe:
seems way more stable now and working with top right and top left arms
improvement in function traverse_nodes_fixed
improvement in function select_by_template

aeon0:
Closed in favour of #348

PR #330, opened 2021-12-24
Fix auto gold pickup bug

LokarBOA:
after auto gold pickup gets turned off due to stash full it causes an error on next gold sighting because the pickit_type property wasn't set correctly here.


mgleed:
Appreciate you helping out. Fixed in #329 :)

LokarBOA:
gleed beat me to it.

PR #333, opened 2021-12-25
Fishymancer

deinhaus:
Added a fishymancer for normal and nightmare (hell is comming soon)
Expanded the buy_pots funtktion for buying GREATER Pods in Nightmare


src/ui/ui_manager.py
```diff
@@ -586,14 +586,23 @@ def buy_pots(self, healing_pots: int = 0, mana_pots: int = 0):
...
-        h_pot = self._template_finder.search_and_wait("SUPER_HEALING_POTION", roi=self._config.ui_roi["vendor_stash"], time_out=3)
+        if self._config.general["difficulty"].upper() == 'NIGHTMARE':
+            healingpotion = "GREATER_HEALING_POTION"
+            manapotion = "GREATER_MANA_POTION"
+        elif self._config.general["difficulty"].upper() == 'HELL':
+             healingpotion = "SUPER_HEALING_POTION"
+             manapotion = "SUPER_MANA_POTION"
+        else:
+             healingpotion = "HEALING_POTION"
+             manapotion = "MANA_POTION"
```
aeon0, line 597: I think there are no templates for "normal" potions yet. Did you maybe forget to add them?

config/params.ini
```diff
@@ -4,23 +4,23 @@ name=Botty
 monitor=0
 custom_message_hook=
 logg_lvl=debug
-max_game_length_s=380
+max_game_length_s=100
```
aeon0, line 7: Please leave params in param.ini to default unless it is necessary and makes sense for all usecases.

aeon0 (changes requested):
Left some comments.

config/params.ini
deinhaus: changed back to default

src/ui/ui_manager.py
deinhaus: deleted this entry and just alowed super and greater potions.

aeon0:
So, there are two PRs implementing a similar feature with #337 

#337 seems to have a bit more already? Regarding the potion buying. We can merge that if you make a separate PR for it.

aeon0:
Clsoing in favour of #337

PR #335, opened 2021-12-27
Blizz sorc Update

D2RLegit:
Faster runs


aeon0:
Closed in favour of #335

PR #337, opened 2021-12-27
Necro class

nlapinski:
adds a basic summon necro template for eldrich,shenk,pindle runs
-no teleport  (only walking, I have no gear to test tele)
-no trav/nihlithak support
-works best if eldritch is run first always
-tested in hell (clvl 80+)
-needs a merc to help first clear for corpse explosion


config/params.ini
```diff
@@ -1,6 +1,6 @@
 ; There is detailed documentation for each parameter in the README.md
 [general]
-name=Botty
```
egut125, line 3: Don't change param defaults without really meaning to. Revert this file to original defaults and only add new fields

config/pickit.ini
```diff
@@ -35,12 +35,12 @@
 ;==================================================================;
 
 ;potions
-misc_rejuvenation_potion=0
+misc_rejuvenation_potion=1
```
egut125, line 38: Don't change param defaults without really meaning to. Revert this file to original defaults and only add new fields

src/bot.py
```diff
@@ -195,12 +199,13 @@ def on_create_game(self):
                 Logger.info(f"Please Enter the region ip and hot ip on config to use")
 
         # Run /nopickup command to avoid picking up stuff on accident
-        if not self._ran_no_pickup:
-            self._ran_no_pickup = True
-            if self._ui_manager.enable_no_pickup():
-                Logger.info("Activated /nopickup")
-            else:
-                Logger.error("Failed to detect if /nopickup command was applied or not")
+        self._ran_no_pickup = True
```
egut125, line 202: Seems very much like unrelated change. Don't include changes that are not necessary for your feature.

egut125 (changes requested):

egut125:
Left some comments

nlapinski:
Reverted the default config files to match the current release. Fixed the accidentally change in bot.py

Added a bunch of stuff to improve shenk+eld runs, and more summoning setup

src/char/necro.py
```diff
@@ -0,0 +1,645 @@
...
+import mouse as _mouse
+from mouse import _winmouse
+
+class Necro(IChar):
+    def __init__(self, skill_hotkeys, char_config, screen: Screen, template_finder: TemplateFinder, ui_manager: UiManager, pather: Pather):
+        Logger.info("Setting up Necro")
+        super().__init__(skill_hotkeys, char_config, screen, template_finder, ui_manager)
+        self._pather = pather
+        #custom pindle paths for necro - there must be a better way?
+        self._pather._paths = {
```
aeon0, line 26: I agree, there must be a better way :D So I assume for some pathes you want to have a different order of nodes? First of all I dont think you want to change any of the town pathing, right? So doing sth like this:
```python
self._pather._paths[Location.A5_PINDLE_START, Location.A5_PINDLE_SAFE_DIST] = [100, 101]
``` 
would be enough. You could add a setter for it inorder to not access private members. similar to the offset_node() function in pather. Sth like `adapt_path(key: tuple(Location, Location), new_node_sequence: list[in])`

nlapinski: added a function to do this in pather.py and replace only the needed paths now

src/char/necro.py
```diff
@@ -0,0 +1,524 @@
...
+from utils.misc import wait, rotate_vec, unit_vector
+import random
+from typing import Tuple, Union, List
+from pather import Location, Pather
+import numpy as np
+import time
+from utils.misc import cut_roi, is_in_roi
+
+import mouse as _mouse
+from mouse import _winmouse
```
aeon0, line 18: unused imports

README.md
```diff
@@ -159,6 +159,20 @@ run_shenk=0
...
+| bone_armor     | Required Hotkey for Bone Armor                                                      |
+| clay_golem     | Required Hotkey for Clay Golem                                                      |
+| raise_skeleton | Required Hotkey for Raise Skeleton                                                  |
+| amp_dmg        | Required Hotkey for Amplify Damage                                                  |
+| corpse_explosion | Required Hotkey Corpse Explosion                                                  |
+| raise_revive   | Required Hotkey revive                                                              |
+| clear_pindle_packs | clears mobs before pindle                                                       |
+
+
+
```
aeon0, line 175:
```suggestion
```

src/bot.py
```diff
@@ -28,6 +28,8 @@
 from char.trapsin import Trapsin
 from char.hammerdin import Hammerdin
 from char.barbarian import Barbarian
+from char.necro import Necro
+
```
aeon0, line 32:
```suggestion
```

src/config.py
```diff
@@ -149,6 +149,12 @@ def __init__(self, print_warnings: bool = False):
         if "trapsin" in self._custom:
             self.trapsin.update(self._custom["trapsin"])
 
+        # Necro config
+        self.necro = self._config["necro"]
+        if "necro" in self._custom:
+            self.necro.update(self._custom["necro"])
+
+
```
aeon0, line 157:
```suggestion
```

src/char/necro.py
```diff
@@ -0,0 +1,524 @@
...
+            x = cast_pos_abs[0] + (random.random() * 2*spray - spray)
+            y = cast_pos_abs[1] + (random.random() * 2*spray - spray)
+            cast_pos_monitor = self._screen.convert_abs_to_monitor((x, y))
+
+            nx = cast_pos_monitor[0]
+            ny = cast_pos_monitor[1]
+            if(nx>1280):
+                nx=1275
+            if(ny>720):
+                ny=715
```
aeon0, line 55: if you are concerned about moving out of the window you should also check for < 0, no?

aeon0 (changes requested):
Some small request changes. Didnt go through the necro class line by line. But outside looks good. I think for your pathing thing that's as good as we got for now : / 

One problem we get with a teleporting necro is, that he will not use the pather pathes but static pathes. So for now this will probably only work for walking necros. But that's fine for me. We can adapt the teleporting part also later if you want to merge.

src/char/necro.py
nlapinski: Yea I'll do another set of changes for static paths (teleport) once I get the gear for it setup. Mostly intended for walking, ill add a readme note on it  and in the parms. I'll take care of these changes then  should be about it!

src/char/necro.py
nlapinski: yep should check that as well, ill take a look and see if this is even still needed.

nlapinski:
Rebased my changes over the latest master to make sure nothing broke, everything seemed to work as expected.

src/config.py
```diff
@@ -157,6 +157,11 @@ def __init__(self, print_warnings: bool = False):
         if "barbarian" in self._custom:
             self.barbarian.update(self._custom["barbarian"])
 
+        # Necro config
+        self.necro = self._config["necro"]
+        if "necro" in self._custom:
+            self.necro.update(self._custom["necro"])
```
aeon0, line 163: With the clear_pindle_pack being a bool and not a string you have to type cast it here:
`self.necro["clear_pindle_pack"] = bool(int(self.necro["clear_pindle_pack"]))`

aeon0 (changes requested):

src/config.py
nlapinski: Added a cast in the necro class, since I'm using the hotkey config (which are strings) it didnt make sense to deal with casting all of them.
The global configs all get casted, but the fX keys just stay strings it looks like.
I could move this to the global config but since its only used by the necro that didnt make sense either.

aeon0: you could leave it as is. So everything is strings, and then only add the line I posted, which will only cast the one field. But fine like this for now.

aeon0 (approved):

PR #338, opened 2021-12-27
Repair Each Run

skizotenigma:
this will have you repair at the start of each run.
not recommended for high runewords.
mainly for poor folk with tele necklaces that have charges.

usage:
set 'always_repair=1' in params.ini

### DIFF
--- a/config/params.ini
+++ b/config/params.ini
@@ -75,9 +75,7 @@ belt_rejuv_columns=2
 belt_hp_columns=1
 belt_mp_columns=1
 pre_buff_every_run=0
-#skizot
 always_repair=0
-#end skizot
 
 ; ===========================
 ; ==== Builds: Sorceress ====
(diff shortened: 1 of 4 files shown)


config/params.ini
```diff
@@ -75,7 +75,9 @@ belt_rejuv_columns=2
 belt_hp_columns=1
 belt_mp_columns=1
 pre_buff_every_run=0
-
+#skizot
```
aeon0, line 78: comments about user information of changes can be removed. That's why we have git : )

aeon0 (changes requested):
Please add a short documentation to the README.md
Otherwise good!

config/params.ini
skizotenigma: lol it's an OLD habit.

aeon0 (approved):

PR #347, opened 2021-12-29
CTA check to make sure you can BO/BC

skizotenigma:


skizotenigma:
bo

bo icon goes in assets\npc\action_btn\

PR #348, opened 2021-12-29
A2 Pathing + Arcane run + Static for Sorc + Static Tele robustification + Chest class

aeon0:
Changes from @kozfelipe I just created the branch.


build.py
```diff
@@ -77,4 +77,4 @@ def clean_up():
 
     if new_version_code is not None:
         os.system(f'git add .')
-        os.system(f'git commit -m "Bump version to v{args.version}"')
```
aeon0, line 80: Dont remove empty lines at end of file

config/params.ini
```diff
@@ -19,12 +19,13 @@ saved_games_folder=
 
 [routes]
 run_trav=0
-run_pindle=1
-run_eldritch=1
+run_pindle=0
+run_eldritch=0
 run_shenk=0
 run_nihlatak=0
+run_arcane=1
```
aeon0, line 26: Leave the default. If a user wants to run any other they will adapt it on their own or use their custom.ini

development.md
```diff
@@ -77,4 +77,4 @@ If you installed your miniconda in another location you will of course have ot c
 # Adapt new version with x.x.x, build .exe and bundeling all needed resource into one folder
 python release.py x.x.x
 ```
-For changelog run: `git log <PREVIOUS_TAG>..HEAD --oneline --decorate`
```
aeon0, line 80: dont remove empty line

src/char/i_char.py
```diff
@@ -42,7 +42,8 @@ def select_by_template(
         template_type:  Union[str, List[str]],
         success_func: Callable = None,
         time_out: float = 8,
-        threshold: float = 0.68
+        threshold: float = 0.68,
+        telekinesis = False
```
aeon0, line 46: typehint as bool `telekinesis: bool = False`

src/char/sorceress/sorceress.py
```diff
@@ -32,6 +33,40 @@ def pick_up_item(self, pos: Tuple[float, float], item_name: str = None, prev_cas
...
         else:
             return super().pick_up_item(pos, item_name, prev_cast_start)
+            
+    def select_by_template(
+        self,
+        template_type:  Union[str, List[str]],
+        success_func: Callable = None,
+        time_out: float = 8,
+        threshold: float = 0.68,
+        telekinesis = False
```
aeon0, line 43: adapt same as in i_char.py

src/char/sorceress/sorceress.py
```diff
@@ -32,6 +33,40 @@ def pick_up_item(self, pos: Tuple[float, float], item_name: str = None, prev_cas
...
+        if type(template_type) == list and "A5_STASH" in template_type:
+            # sometimes waypoint is opened and stash not found because of that, check for that
+            if self._template_finder.search("WAYPOINT_MENU", self._screen.grab()).valid:
+                keyboard.send("esc")
+        start = time.time()
+        while time_out is None or (time.time() - start) < time_out:
+            template_match = self._template_finder.search(template_type, self._screen.grab(), threshold=threshold)
+            if template_match.valid:
+                x_m, y_m = self._screen.convert_screen_to_monitor(template_match.position)
+                if self._skill_hotkeys["telekinesis"] and telekinesis:
```
aeon0, line 54: Instead of a another nested if here you could check at the very beginning like this:
```python
if not (self._skill_hotkeys["telekinesis"] and telekinesis):
  return super().select_by_template(tempalte_type, success_func, time_out, threshold)
# Implement telekinsis specific implementation
...
```

src/char/trapsin.py
```diff
@@ -149,5 +149,4 @@ def kill_nihlatak(self, end_nodes: list[int]) -> bool:
     t_finder = TemplateFinder(screen)
     pather = Pather(screen, t_finder)
     ui_manager = UiManager(screen, t_finder)
-    char = Trapsin(config.trapsin, config.char, screen, t_finder, ui_manager, pather)
-
```
aeon0, line 153: Empty line removed

src/chest.py
```diff
@@ -0,0 +1,88 @@
...
+            if template_match.name is None:
+                break
+            if score <= threshold:
+                templates.remove(template_match.name)
+                continue
+            if template_match.valid:
+                Logger.debug(f"Opening {template_match.name} ({template_match.score*100:.1f}% confidence)")
+                x_m, y_m = self._screen.convert_screen_to_monitor(template_match.position)
+                # act as picking up a potion to support telekinesis
+                self._char.pick_up_item([x_m, y_m], 'potion')
```
aeon0, line 45: We might want to rename this, but for now the workaround is fine. Maybe just add a TODO comment to the pick_up_item function in i_char. e.g. `# TODO: Rename function as it is used for general object pick up / selection`.

src/pather.py
```diff
@@ -142,6 +161,17 @@ def __init__(self, screen: Screen, template_finder: TemplateFinder):
...
             302: {"TRAV_V3_0": (-18, 6), "TRAV_V3_7": (73, 208), "TRAV_V3_8": (339, 133), "TRAV_V3_6": (65, 398), "TRAV_V3_5": (-440, 146)},
             304: {"TRAV_V2_4": (125, -148), "TRAV_V2_3": (-187, 55), "TRAV_V2_1": (-207, 59), "TRAV_V2_2": (267, 183), "TRAV_V2_0": (-159, 403)},
+            # Arcane
+            450: {"ARC_START": (49, 62)},
+            451: {"ARC_PLATFORM_1": (0, 0), "ARC_PLATFORM_2": (125, -70), "ARC_CENTER": (131, -78)},
+            452: {"ARC_PLATFORM_2": (-249, 69), "ARC_CENTER": (-242, 62), "ARC_END_1": (122, 73), "ARC_END_2": (-245, 76), "ARC_END_3": (-248, -136)},
+            453: {"ARC_START": (-259, 62)},
+            454: {"ARC_PLATFORM_1": (237, 72), "ARC_PLATFORM_3": (210, 89), "ARC_CENTER": (237, 72), "ARC_END_1": (601, 82), "ARC_END_2": (234, 86), "ARC_END_3": (232, -127)},
+            456: {"ARC_START": (145, 264)},
+            457: {"ARC_PLATFORM_3": (-174, -103), "ARC_CENTER": (-146, -120), "ARC_END_1": (218, -109), "ARC_END_2": (-307, -59)},
```
kozfelipe, line 171: 458 was deprecated, should we re-index?

src/chest.py
```diff
@@ -0,0 +1,88 @@
...
+            if filename.endswith('.png'):
+                chest = filename[:-4].upper()
+                self._templates.append(chest)
+                
+    def open_up_chest(
+        self,  
+        time_out: float = 8,
+        threshold: float = 0.8
+    ) -> float:
+        score = 100
```
aeon0, line 31: template match scores are in float range of [0, 1]. That 100 confuses me.

src/chest.py
```diff
@@ -0,0 +1,88 @@
...
+        time_out: float = 8,
+        threshold: float = 0.75
+    ) -> bool:
+        # keep opening chests till no matches nearby
+        score = 100
+        while score >= threshold:
+            score = self.open_up_chest(time_out, threshold)
+        score = 100
+        while score >= threshold:
+            score = self.open_up_chest(time_out, threshold)
```
aeon0, line 67: Why same code twice?

src/run/pindle.py
```diff
@@ -38,7 +38,7 @@ def approach(self, start_loc: Location) -> Union[bool, Location]:
             return False
         wait(0.5, 0.6)
         found_loading_screen_func = lambda: self._ui_manager.wait_for_loading_screen(2.0)
-        if not self._char.select_by_template(["A5_RED_PORTAL", "A5_RED_PORTAL_TEXT"], found_loading_screen_func):
+        if not self._char.select_by_template(["A5_RED_PORTAL", "A5_RED_PORTAL_TEXT"], found_loading_screen_func, telekinesis=True):
```
aeon0, line 41: Specifically this we can not do, because using telekinesis will make the sorc end up on a different starting position than when entering the portal by foot. Then the static pathing is all messed up.

src/char/sorceress/sorceress.py
```diff
@@ -32,6 +33,40 @@ def pick_up_item(self, pos: Tuple[float, float], item_name: str = None, prev_cas
             return cast_start
         else:
             return super().pick_up_item(pos, item_name, prev_cast_start)
+            
+    def select_by_template(
```
aeon0, line 37: Is there a fail save in case telekensis is out of range? We will just try tele forever / until timeout now, right?

src/town/town_manager.py
```diff
@@ -40,7 +43,8 @@ def wait_for_town_spawn(self, time_out: float = None) -> Location:
         template_match = self._template_finder.search_and_wait([
             "A5_TOWN_0", "A5_TOWN_1",
             "A4_TOWN_4", "A4_TOWN_5",
-            "A3_TOWN_0", "A3_TOWN_1"
+            "A3_TOWN_0", "A3_TOWN_1",
+            "A2_TOWN_LIGHT_WELL", "A2_TOWN_MEDIUM_WELL", "A2_TOWN_DARK_WELL"
```
aeon0, line 47: Is it all the same template just on different daytimes? That is a bit tricky. Cause it can happen that your merc is in front of that template. That is why all the other town templates are from 2 different locations.

src/town/a2.py
```diff
@@ -0,0 +1,62 @@
...
+        wait(0.5, 0.7)
+        found_wp_func = lambda: self._template_finder.search("WAYPOINT_MENU", self._screen.grab()).valid
+        return self._char.select_by_template(["A2_WP_LIGHT", "A2_WP_DARK"], found_wp_func, telekinesis=True)
+
+    def wait_for_tp(self) -> Union[Location, bool]:
+        template_match = self._template_finder.search_and_wait("A3_TOWN_10", time_out=20)
+        if template_match.valid:
+            self._pather.traverse_nodes((Location.A3_STASH_WP, Location.A3_STASH_WP), self._char, force_move=True)
+            return Location.A3_STASH_WP
+        return False
```
aeon0, line 62: Need to be adapted to A2. What the function does is: "Wait that we are back in A2 at town portal location". Since this location is not the same then the start position in all acts there is a separate function for it. Since this is the case for A2, you will have to also traverse to a known point. For A2 traversing from town portal -> Location.A2_FARA_STASH probably makes the most sense (and returning this location). From that point botty can pick up any new action.

src/run/arcane.py
```diff
@@ -0,0 +1,213 @@
...
+        self._town_manager = town_manager
+        self._ui_manager = ui_manager
+        self._char = char
+        self._pickit = pickit
+        self._chest = Chest(self._char, self._template_finder, 'arcane')
+    
+    def approach(self, start_loc: Location) -> Union[bool, Location]:
+        Logger.info("Run Arcane")
+        #if not self._char.can_teleport():
+        #    raise ValueError("Arcane requires teleport")
```
aeon0, line 35: this should not be commented

aeon0 (commented):
The only thing I havent taken a deeper look into is the arcane battle() function. Want to do this while actually testing. 
Otherwise good job already!

src/run/pindle.py
kozfelipe: I tested in act5 and works flawless

src/town/town_manager.py
kozfelipe: Its a really odd situation mercenary in front of the well. Even so, wait_for_town_spawn() has a high timeout value enought to mercenary leave that area

development.md
kozfelipe: weird I didnt even touch this file

src/chest.py
kozfelipe: I think we might replace this with select_by_template(telekinesis=true)

src/chest.py
kozfelipe: will replace it for 1

src/chest.py
kozfelipe: We get duplicated versions of chests in the arcane endpoint very often.  If we get a locked chest without any keys left, it will remove the template from the list to avoid endless loop through the locked chest and the duplicated chest wont be clicked. Sometimes it doenst recognize the chest in the first attempt, but it does in the second after opening the others

src/char/sorceress/sorceress.py
kozfelipe: I havent seen any cases out of range

src/town/a2.py
kozfelipe: added tp to wp path

PR #349, opened 2021-12-30
Backup and Restore settings

FrancescoZini:
Add the capability to backup and restore D2R settings.

Also refactored the main.py to extract game logic out of the menu with a new class: GameController:
This class will be the main controller thread of the bot that will take care of handling the bot itself and is separate from the program main-loop used to listen to keyboard events.

The menu is now using the python keyboard built-in functions to handle the main loop rather than relying on a custom-built one.

EDIT0:
Also refactored the graphic debugger to extract its logic in a separate thread handled by the GraphicDebuggerController.
The main now has both GameController and DebuggerController and they handle the events to start and stop themselves as an Aut Aut, only one of them can run at a time.

More specifically with regards of the keys, now they will behave as follow:
F11 will run botty if  botty is not running and the graphic debugger is not running
F11 will kill the graphic debugger and run botty if the graphic debugger was running
F11 will pause botty if botty was already running

F10 will start the debugger if the grapchi debugger and botty are not running
F10 will stop the graphic debugger if it was running
F10 will kill botty and start the graphic debugger if botty was running


src/game_controller.py
```diff
@@ -0,0 +1,99 @@
...
+from game_stats import GameStats
+from health_manager import HealthManager
+from logger import Logger
+from messenger import Messenger
+from screen import Screen
+from utils.misc import kill_thread
+
+
+class GameController:
+
```
aeon0, line 21: remove new line

src/game_controller.py
```diff
@@ -0,0 +1,99 @@
...
+        if not self.death_manager:
+            # Run death monitor thread
+            self.death_manager = DeathManager(self.screen)
+            death_monitor_thread = threading.Thread(target=self.death_manager.start_monitor)
+            death_monitor_thread.daemon = True
+            death_monitor_thread.start()
+        if not self.game_recovery:
+            self.game_recovery = GameRecovery(self.screen, self.death_manager)
+        if not self.game_stats:
+            self.game_stats = GameStats()
```
aeon0, line 49: All the object creation could be done in the __init__() without these if clauses, no? Adding the config as member var and ditching it from the run_bot() method.

src/game_controller.py
```diff
@@ -0,0 +1,99 @@
...
+            max_game_length_reached = self.game_stats.get_current_game_length() > config.general["max_game_length_s"]
+            if max_game_length_reached or self.death_manager.died() or self.health_manager.did_chicken():
+                # Some debug and logging
+                if max_game_length_reached:
+                    Logger.info(f"Max game length reached. Attempting to restart {config.general['name']}!")
+                    if config.general["info_screenshots"]:
+                        cv2.imwrite(
+                            "./info_screenshots/info_max_game_length_reached_" + time.strftime(
+                                "%Y%m%d_%H%M%S") + ".png",
+                            bot._screen.grab())
```
aeon0, line 73: I understand that some ppl have ocd for line length. But this could maybe be moved to L72? ;)

src/utils/auto_settings.py
```diff
@@ -31,6 +24,39 @@ def adjust_settings():
...
+            print("Settings backed up successfully.")
+    else:
+        shutil.copyfile(d2_saved_games + "\\Settings.json", d2_saved_games + "\\Settings_backup.json")
+        print("Settings backed up successfully.")
+
+
+def restore_settings_from_backup(config: Config):
+    close_down_d2()
+    d2_saved_games = get_d2r_folder(config)
+    shutil.copyfile(d2_saved_games + "\\Settings_backup.json", d2_saved_games + "\\Settings.json")
```
aeon0, line 45: This will fail with a "FileNotFoundError" if the user did not back up a settings file previously. Maybe catch the error and output some info to the user instead and return.

src/main.py
```diff
@@ -108,31 +51,14 @@ def main():
...
-        elif keyboard.is_pressed(config.general['graphic_debugger_key']):
-            run_graphic_debugger()
-            break
-        time.sleep(0.02)
+    keyboard.add_hotkey(config.general['auto_settings_key'], lambda: adjust_settings(config))
+    keyboard.add_hotkey(config.general['graphic_debugger_key'], lambda: run_graphic_debugger())
+    keyboard.add_hotkey(config.general['restore_settings_from_backup_key'], lambda: restore_settings_from_backup(config))
+    keyboard.add_hotkey(config.general['settings_backup_key'], lambda: backup_settings(config))
+    keyboard.add_hotkey(config.general['resume_key'], lambda c: start_bot(c), args=[config])
+    keyboard.add_hotkey(config.general["exit_key"], lambda: Logger.info(f'Force Exit') or os._exit(1))
```
aeon0, line 59: why I did not do it this way is to avoid having graphic debugger run while botty is running and to not have multiple bottys run at the same time. The logic now is that the "resume_key" is overwritten within run_bot() and thus will not trigger another start_bot() as I see it. A bit hacky, but good enough for me.

src/main.py
```diff
@@ -108,31 +51,14 @@ def main():
...
-            break
-        if keyboard.is_pressed(config.general['auto_settings_key']):
-            adjust_settings()
-            break
-        elif keyboard.is_pressed(config.general['graphic_debugger_key']):
-            run_graphic_debugger()
-            break
-        time.sleep(0.02)
+    keyboard.add_hotkey(config.general['auto_settings_key'], lambda: adjust_settings(config))
+    keyboard.add_hotkey(config.general['graphic_debugger_key'], lambda: run_graphic_debugger())
```
aeon0, line 55: I can not exit the graphic debugger anymore with f12. Not sure how to resolve this. Seems like cv2.waitKey() is blocking keyboard? Can you check if you have the same behaviour and look into this?

aeon0 (changes requested):
Nice refactoring and helpful feature. Also quality code, so I was a bit more picky in the review than usually around here : )

src/game_controller.py
FrancescoZini: yeah very fair point I did it actually before reading your comment, this was realyl part of mid changes I made to transition the logic into the new classes and should definetely go in the constructor

src/game_controller.py
FrancescoZini: Done, I've also made it access the self.screen rather than bot screen as it is the same.
I'm using PyCharm I set the line wrap at 180 chars instead of 120 for this project.

src/main.py
FrancescoZini: ahhh right, it is because the run_graphic_debugger is not a strictly self-contained function, that means it does its own loop (while 1:...) inside it so on the invokation is not return the keyboard control to the main-loop.

I'll look into this, a solution would be to have the graphic debugger on a separate thread with its own loop (following the same pattern I used for splitting the game controller logic). That would also allow to stop the graphic debugger without killing botty.

FrancescoZini: reading the other comment I now understand better the initial logic behind the initial code, basically you want the following:

run EITHER botty OR graphic_debugger.
Specifically prevent the situation in which you might have both running at the same time.

is that correct?

src/game_controller.py
FrancescoZini: Done

src/utils/auto_settings.py
FrancescoZini: This is now handling the non existent file. I personally prefer not code around try/catch so I added a check that the path exists before attempting to copy. Tested locally seems to work fine

src/main.py
FrancescoZini: Fair enough, I now better understood your intention with the first code iteration and I agree that it felt hacky to re-bound the key, so I changed it to let the game controller take care of that. So if botty is running and you want to pause it, F11 will not pause the bot, if it is not running it will start it instead. That without removing the hotkey and re-adding it

src/main.py
FrancescoZini: I made some more changes to make the wanted behavior work , id est:

F11 will run botty if  botty is not running and the graphic debugger is not running
F11 will kill the graphic debugger and run botty if the graphic debugger was running
F11 will pause botty if botty was already running

F10 will start the debugger if the grapchi debugger and botty are not running
F10 will stop the graphic debugger if it was running
F10 will kill botty and start the graphic debugger if botty was running

aeon0:
GJ!

aeon0 (approved):

PR #352, opened 2021-12-30
"Basic" classes

skizotenigma:
will left attack unless right attack is available.
"basic" = just a basic melee range character (nova etc)
"basic_ranged" = just a basic ranged character (fireball etc)

these can be used with any class / build, provided the left and right attacks are both useable (auras cause issues)
you have your left and right attack, and two buffs you can set (for prebuffing)


src/char/basic_ranged.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+class Basic_Ranged(IChar):
+    def __init__(self, skill_hotkeys, char_config, screen: Screen, template_finder: TemplateFinder, ui_manager: UiManager, pather: Pather):
+        Logger.info("Setting up Basic Ranged Character")
+        super().__init__(skill_hotkeys, char_config, screen, template_finder, ui_manager)
+        self._pather = pather
+        self._do_pre_move = True
+        if not self._skill_hotkeys["teleport"]:
+            self._do_pre_move = False
```
aeon0, line 23: I dont think we need this here. This is just specifically for walking pala to switch from vigor to concentration for the last node traverse as we already attack quite a bit. Such a specific thing shouldn't be done here imho.

src/char/basic_ranged.py
```diff
@@ -0,0 +1,216 @@
...
+        keyboard.send(self._skill_hotkeys["right_attack"])
+        for _ in range(3):
+            x = cast_pos_abs[0] + (random.random() * 2 * spray - spray)
+            y = cast_pos_abs[1] + (random.random() * 2 * spray - spray)
+            cast_pos_monitor = self._screen.convert_abs_to_monitor((x, y))
+            mouse.move(*cast_pos_monitor)
+            mouse.click(button="right")
+            
+    def pre_buff(self):
+        keyboard.send(self._char_config["stand_still"], do_release=False)
```
aeon0, line 47: Any reason for standstill when using right click skills?

src/char/basic.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+class Basic(IChar):
+    def __init__(self, skill_hotkeys, char_config, screen: Screen, template_finder: TemplateFinder, ui_manager: UiManager, pather: Pather):
+        Logger.info("Setting up Basic Character")
+        super().__init__(skill_hotkeys, char_config, screen, template_finder, ui_manager)
+        self._pather = pather
+        self._do_pre_move = True
+        # offset shenk final position further to the right and bottom
+        self._pather.offset_node(149, [120, 70])
```
aeon0, line 21: I would offset this only if teleport is available. It is chosen that way because of all the mobs its hard to reach that point without teleport.

src/char/basic.py
```diff
@@ -0,0 +1,157 @@
...
+        # In case we have a running barb, we want to switch to ???? when moving to the boss
+        # as most likely we will click on some mobs 
+        if not self._skill_hotkeys["teleport"]:
+            self._do_pre_move = False
```
aeon0, line 25: No need for that. Sth specific for walking pala

aeon0 (changes requested):

aeon0 (approved):

PR #358, opened 2021-12-31
Enhanced discord webhook.

skizotenigma:
Makes the discord purtier.


skizotenigma:
eguts is way better. refactoring mine off his base.

PR #359, opened 2021-12-31
Bug Fix: F11 pressed repeatedly might crash the main loop

FrancescoZini:
The safest solution would be to try and catch all in the main loop but that would also most likely mask exceptions that we might want to make the program crash, hence would probably be an overkill.

Checking that self.bot is initialized in the game controller will fix this issue.


src/game_controller.py
```diff
@@ -44,7 +44,6 @@ def run_bot(self, pick_corpse: bool = False):
         self.health_manager.set_callback(lambda: self.bot.stop() or kill_thread(self.bot_thread))
         self.health_manager.set_belt_manager(self.bot.get_belt_manager())
         do_restart = False
-        keyboard.add_hotkey(self._config.general['resume_key'], lambda: self.bot.toggle_pause())
```
aeon0, line 47: But removing this will not allow for any more pausing, right?

aeon0 (changes requested):

src/game_controller.py
aeon0: Ah no, because we have the start_or_pause() function which also does pausing.

aeon0 (approved):

PR #360, opened 2022-01-01
Create run.bat

JohnStyleZ:
run bat instead of typing commands in cmd


aeon0:
Kind of spoon feeding... but ok : )

PR #365, opened 2022-01-01
Fix for crash on A2 Summoner for Sorceress

hectorgimenez:
I was getting a crash when trying A2 summoner with my blizz sorc, looks like the right_attack and left_attack methods are only defined on Trapsin class and not for the Sorceress base class used by light and blizz ones.

I replaced the calls by the same strategy used for pindleskin and it works, probably not the best one but working and not crashing.

### DIFF
--- a/src/char/sorceress/light_sorc.py
+++ b/src/char/sorceress/light_sorc.py
@@ -147,13 +147,12 @@ class LightSorc(Sorceress):
 
     def kill_summoner(self) -> bool:
         # Attack
-        delay = [0.2, 0.3]
         cast_pos_abs = np.array([0, 0])
         pos_m = self._screen.convert_abs_to_monitor((-20, 20))
         mouse.move(*pos_m, randomize=80, delay_factor=[0.5, 0.7])
         for _ in range(int(self._char_config["atk_len_arc"])):
-            self._right_attack(cast_pos_abs, delay, 11)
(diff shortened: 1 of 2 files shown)


aeon0:
Ah, sry. I think I was just one or two minutes quicker : /

aeon0:
But yes, that is totaly the fix. Good catch, thx anyway!

src/char/sorceress/light_sorc.py
```diff
@@ -147,13 +147,12 @@ def kill_nihlatak(self, end_nodes: list[int]) -> bool:
...
         # Attack
-        delay = [0.2, 0.3]
         cast_pos_abs = np.array([0, 0])
         pos_m = self._screen.convert_abs_to_monitor((-20, 20))
         mouse.move(*pos_m, randomize=80, delay_factor=[0.5, 0.7])
         for _ in range(int(self._char_config["atk_len_arc"])):
-            self._chain_lightning(cast_pos_abs, delay, 11)
-            self._lightning(cast_pos_abs, delay, 11)
+            self._chain_lightning(cast_pos_abs, spray=11)
+        self._lightning(cast_pos_abs, spray=50)
```
aeon0, line 155:
```suggestion
            self._lightning(cast_pos_abs, spray=50)
```

PR #370, opened 2022-01-01
Breaking Messenger Module into something more flexible for specif…

egut125:
…ic apis


src/main_copy.py
```diff
@@ -0,0 +1,87 @@
+import keyboard
```
aeon0, line 1: remove please

src/api/generic_api.py
```diff
@@ -0,0 +1,59 @@
+from dataclasses import dataclass
+from config import Config
+import json
+import requests
+
+class GenericApi:
+    def __init__(self):
+        self._config = Config()
+
+    def send(self, msgData):
```
aeon0, line 10: I am not a massive fan of the msgData param where it is a dict and depending on the "type" it has different dict fields. Kind of error prone.

src/api/generic_api.py
```diff
@@ -0,0 +1,59 @@
+from dataclasses import dataclass
+from config import Config
+import json
+import requests
+
+class GenericApi:
```
aeon0, line 6: The "correct" way would be to create an api interface class (class IApi) which implements all the methods that should be implemented by as we have it e.g. with i_char.py. 
I mean, I do not expect that we have many more API interface implementations. Still I would like to have it done properly.

aeon0 (changes requested):
A few concerns about the code architecture.

src/api/generic_api.py
egut125: Any other ways youd like to see? I'm struggling to think of a good one.  The goal was to keep everything going through messenger.send so the rest of botty just has to go through that function.  I guess thats kind of pointless if they need a type field though, so should the send message just be split out into a method for each of the different types?

aeon0: you could use something like @dataclass and then have a union of options. But I would be ok with the dict, just add a comment that explains the logic. So when a developer wants to use the function knows what to put into msgData.

egut125: I went with breaking it down into separate functions.  I post an update with the interfaces and new functions.  Let me know what you think?

https://github.com/aeon0/botty/pull/370/commits/676ac1fb5f9a10608d3edd80c40e376f310e2cc8

egut125:
Ready to be checked again
@aeon0

aeon0:
Now I like it!

PR #371, opened 2022-01-01
Updating Beautiful Table Styling

egut125:
image

Preview included


aeon0:
Even more beautiful? 💅

PR #374, opened 2022-01-02
Restart game 

Master-Matt:
This PR aims to make botty robust against all forms of game interruption.

I tested 3 main conditions:

-  D2R crashes while botty is running.

The restart module now reopens the game and resets the screen so that it should work in the new position

- An error happens that brings you back to main menu

The main testing I did for this was to click on offline tab between games or right after d2r boots up.
The code I left in will always select the non-offline tab before continuing but the offline line is in there but commented out

- D2R was not started before you ran botty.

As a side-effect of the way I implemented restart, you can now just start botty and it will start d2r after the first max game length times out. If we like that, it can be changed so that it checks if d2r is running and starts it.

I haven't added any coverage tests and I'd like some guidance on how you think I should do that. So far things have been running good for me, but I would like more testing to be sure, as this does affect the control flow of botty.


aeon0:
Yeah, any testing that involves d2r itself running is pretty though. I don't know a good solution to that...

Master-Matt:
I ran it all night last night and the only thing that seemed off was a lot of max game lengths around Cain and Qual Kek in A5.
Also twice in A3, we would click on cain instead of stash and hit max game length there. for some reason it was not catching the exit screen and going straight to looking for main menu. I will look into that more.

src/ui/ui_manager.py
```diff
@@ -51,10 +51,11 @@ def use_wp(self, act: int, idx: int):
         wait(0.4, 0.5)
         mouse.click(button="left")
         # wait till loading screen is over
-        if self.wait_for_loading_screen():
+        if self.wait_for_loading_screen(3):
```
Master-Matt, line 54: This was causing a game timeout by the loop just sitting and waiting for the loading screen. But if you exited the wp menu after opening it (like I'm seeing happen sometimes) it will just sit. This now exits the game

aeon0 (changes requested):
Small change requests. Overall I like it and I think its good.

Master-Matt:
Should be good now. rans a few tests and it all worked.

aeon0 (approved):

src/game_controller.py
```diff
@@ -75,45 +76,71 @@ def run_bot(self, pick_corpse: bool = False):
             if self._config.general["info_screenshots"]:
                 cv2.imwrite("./info_screenshots/info_could_not_recover_" + time.strftime("%Y%m%d_%H%M%S") + ".png", self.screen.grab())
             Logger.error(
-                f"{self._config.general['name']} could not recover from a max game length violation. Shutting down everything.")
+                f"{self._config.general['name']} could not recover from a max game length violation. Restarting the Game.")
+            if self._config.general["custom_message_hook"]:
+                messenger.send(msg=f"{self._config.general['name']}: got stuck and will now restart D2R")
+            if restart_game(self._config.general["d2r_path"]):
```
FrancescoZini, line 82: is it possible to please move this to be an optional feature until we have the ability to select the correct Character from the hero selection screen? 

D2R restarted for me last night and upon restart botty played with my level 8 paladin as it is the first character I have on startup

I believe something like `if self._config.whatever["restart_config_name_whatever"]:` before this line + the addition of the config should suffice

aeon0: #403 Can you test this one? @FrancescoZini

FrancescoZini: wow that was fast! :D yeah looks good, I've just left a comment for the case in which the character is down the list (if you have many) so you might have to scroll down

aeon0: Yeah... but... ahhh

Master-Matt: Yeah, @FrancescoZini, what you experienced is a bug in D2R. It should always default to the last character you used. In the last patch they said they fixed it, but.ive been able to reproduce it by logging in to an account across multiple PC's or vms.

PR #376, opened 2022-01-02
Add ability to override every configuration parameter via environment variables

VladimirMakaev:
I'm running botty from source and prefer to commit my own settings to the param.ini and keep in my forks. However some settings like "custom_message_hook" could be sensitive and I'd prefer to override this on the VM itself

If this PR is merged we'll be able to override any configuration parameter in a form **BOTTY_{SECTION}_{KEY}**. E.g to override "custom_message_hook" we can set 

**BOTTY_GENERAL_CUSTOM_MESSAGE_HOOK**

#### In Powershell
` set-item -Path env:\BOTTY_GENERAL_CUSTOM_MESSAGE_HOOK -Value "http://test.com/test"`

#### In REPL
```
>>> from config import Config
>>> c = Config()
>>> c.general["custom_message_hook"]
'http://test.com/test'
>>> c.general["resume_key"]          
'f11'
```

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -15,6 +15,9 @@ class ItemProps:
 
 class Config:
     def _select_val(self, section: str, key: str = None):
+        value_from_env = os.getenv(f'BOTTY_{section.upper()}_{key.upper() if key is not None else ""}')
+        if value_from_env is not None:
+            return value_from_env
         if section in self._custom and key in self._custom[section]:
             return self._custom[section][key]
         elif section in self._config:


aeon0:
This seems like a redundant feature since you can do this exact thing with a `custom.ini`. There you can overwrite every param at it will not be included to the git.
So if you want to overwrite custom_message_hook you would add this to custom.ini
```ini
[general]
custom_message_hook=my_own_link
```
and botty would take that one.

Are you not happy how this is handled or did you just not know about it? Feedback always welcome.

VladimirMakaev:
I think using custom.ini makes sense. I see it's already in .gitignore so I'll abandon this PR then. Thanks for looking into it

PR #379, opened 2022-01-02
Feature: Allow Classic players to shop hammerdin scepters from Drognan

fiseman:
This was requested on Discord. Figured it would be a good first exercise for me.

Added a new drognan script which is based on the anya shopper script. It moves in and out of town since there's no convenient red portal like anya has. Scepters instead of Claws, and it's pretty much the same thing. The shopper code can definitely benefit from some generalization so more things can be configuration based.

And of course added the necessary images for Drognan and the scepters.

Tested locally and appears to work as requested!


aeon0 (approved):
Good job for the first exercise : ) 
I haven't looked in full detail into the dragon shopper (reviews on shopper are a bit more relaxed), but everything else looks good.

PR #381, opened 2022-01-03
Act 1 town   full usage

skizotenigma:


aeon0 (approved):
You went all out on A1, hmm? :D
Not 100% sure we need A1 vendors much, but ok, better having not having I guess : )

skizotenigma:
> 

well there's always the future, and they may be needed there. :D

PR #382, opened 2022-01-03
set D2R window to be always on top

FrancescoZini:
Set D2R window always on top on Windows (using win32gui library).

Not sure if maybe this should be a new button on the menu instead? Currently runs only at the start of botty or when game_recovery is invoked.


aeon0:
I am not 100% convinced it is always wanted. I think it is good if the user has some time read the console. Imagen you have never started botty before. You press exe it goes to fullscreen d2r window. The user might think: "Well, what the hell am I supposed to do now??"

FrancescoZini:
> 

Fair point. What about having a boolean config then, defaulted to 0 so then only experienced users might want to set it to 1.

In my case I'd leave it always to 1 as I have botty running on a vm but from time to time either the mouse goes outside of the window (don't really understand how it is possible yet) or some other window pop-up and overlap d2r making botty stop, I think this might be a useful feature for those who have a setup similar to mine.

FrancescoZini:
> I am not 100% convinced it is always wanted. I think it is good if the user has some time read the console. Imagen you have never started botty before. You press exe it goes to fullscreen d2r window. The user might think: "Well, what the hell am I supposed to do now??"

What about these new changes I made? 

Moved D2R window always on top to the config under the advanced_options and defaulted to 0.

Now the D2R window is set always on top only when the bot start (F11) and when it pause (F11 again) or F12 is pressed the D2R window (if it still exists) will be restored to its original value (non-topmost)

aeon0 (approved):
I like it

PR #384, opened 2022-01-03
Fixed sorc 301 pathing.

D2RLegit:
Reduces errors and delays top left node in trav.


aeon0 (approved):

src/char/sorceress/blizz_sorc.py
```diff
@@ -137,14 +137,10 @@ def kill_council(self) -> bool:
         pos_m = self._screen.convert_abs_to_monitor((-285, -320))
         self.pre_move()
         self.move(pos_m, force_move=True)
+        wait(0.5)
         # Move to far left
-        #new node top left
+        self._pather.offset_node(301, (-80, -30))
```
aeon0, line 142: You have to be careful here. This will offset the node everytime you run trav. If you want to offset it once you have to do it in the __ini__() function

aeon0 (changes requested):

src/char/sorceress/blizz_sorc.py
D2RLegit: Should be permanent unless causes pathing issues.

D2RLegit (commented):
fixed 301 node offset

aeon0 (approved):

PR #385, opened 2022-01-03
1 liner bug fix for game_recovery: name 'config' is not defined

FrancescoZini:
as per the title


aeon0 (approved):
Guess I didnt review very well : )

FrancescoZini:
> Guess I didnt review very well : )

humble :) but it was my bad really, I'm working with 2 IDEs and somehow I did this change on one and the rest of the changes on the other one so I missed it

PR #386, opened 2022-01-03
Add some items and filters (rare matriarchal jav, +3 tornado pels, +3 warcry helmets)

D2RLegit:


aeon0 (changes requested):
Shako is "uniq_helm_..." so its not missing in current one. 
We renamed all "eth_" to "gray_"

D2RLegit (commented):
ignore blizz_sorc.py changes

PR #389, opened 2022-01-03
Bug fix: import keyboard in game_recovery.py

FrancescoZini:
For some mysterious reason my PyCharm detects this import as unused, that's why it got deleted in the first place.


aeon0:
Did it as a "drive-by-fix" a few minutes ago : ) 
https://github.com/aeon0/botty/pull/388/files

PR #390, opened 2022-01-04
Auto format code with black via pre-commit

reqyl:
This will auto-format all python code consistently with [black](https://github.com/psf/black) via [pre-commit](https://pre-commit.com/) on commit.

- Added black lint to github CI action (check if any file would be changed)
- Added pre-commit with config to format with black, remove trailing whitespace and always end files with newline
- Document how to install the pre-commit hook for development (`pre-commit install`)
- Manually trigger the hook for all files to do the initial formatting (`pre-commit run -a`)

Using black improves collaboration by erasing any inconsistencies in code formatting, letting reviews etc be purely focused around code and functionality.

The git-hook must be installed manually by contributors via `pre-commit install` before the hook is actually run on commit. But the lint task in CI should stop PRs from merging if code is not aligned with black formatting.

Something to consider if/when merging this PR is any current work-in-progress PRs would probably face conflicts and need to run pre-commit on their branch.


reqyl:
Hm a few knobs like line-length can be set with black but it's mostly all or nothing.

Before I tried black I was also skeptical, but after using it I think it's such a relief not having to think about formatting at all.
Everyone has different styles and preferences so a formatter will never fully align, but the style is decent and **consistent** which I think is worth more.

Here's a [nice talk](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=524s&ab_channel=TheGoProgrammingLanguage) 
 that mentions the built in formatter of golang and the reasoning behind it. 


>*Yeah I don't like how it formats, but I really like that it formats*
>*Gofmt is no one's favorite, yet gofmt is everyones favorite*

I'm closing this PR but can reopen it if you'd reconsider.

aeon0:
Thanks for the talk. There are some really good points! I will think about it when I have some time on my hands.

PR #391, opened 2022-01-04
Npc serach optimization

aeon0:


environment.yml
```diff
@@ -19,4 +19,4 @@ dependencies:
     - pytest-pythonpath
     - graphviz
     - psutil
-    - pywin32
\ No newline at end of file
+    - pywin32==225
```
FrancescoZini, line 22: @aeon0 by the way, specifying the version 225 seems to be breaking for me. Any particular reason why you specified this version?

I did some more research on this topic and pywin32 and pip don't seem to work well together, I tried to move pywin32 dependency outside of pip (literally between python and pip in this very file. That worked, I tried two clean install in 2 different machines with that change and seems ok, I will create a pull request shortly.

aeon0: It worked for one machine and I have seen some StackOverflow comment that suggested to change to that version for python 3.9
the difference of moving it to pip or outside of it, is that it will be installed with conda. Installing via conda can do more than only installing a python module. It might setup win32api, ship a DLL, etc. But if it works, that's good. I merge your fix pr.

PR #394, opened 2022-01-04
pywin32 dependency fix

FrancescoZini:
Moved pywin32 dependency outside of pip (tested a clean install on two machines with this change and worked fine)

Also removed the win32api import that was causing the issue.

Both changes fix the issue that several people are expereincing on discord separately (either would fix it but both are better).


aeon0 (approved):

aeon0:
That for example does not work for me on one machine... I am a bit lost with this one : / 
It seems different solutions work for different ppl but there is no one solution that works for all...

FrancescoZini:
Interesting... which kind of error are you getting? Runtime when you run botty or during the env creation? If it is something like "module not found" I think our best bet would be to go with the heavy artillery, as they propose on their wiki to run this python Scripts/pywin32_postinstall.py -installhttps://github.com/mhammond/pywin32On 5 Jan 2022, at 20:34, aeon0 ***@***.***> wrote:That for example does not work for me on one machine... I am a bit lost with this one : / It seems different solutions work for different ppl but there is no one solution that works for all...—Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications on the go with GitHub Mobile for iOS or Android. You are receiving this because you authored the thread.Message ID: ***@***.***>

aeon0:
Na, the DLL not found error.

FrancescoZini:
By the way, I'm looking to find some alternatives to this, I was not expecting this library to be so annoying.On 5 Jan 2022, at 21:33, Francesco Zini ***@***.***> wrote:Interesting... which kind of error are you getting? Runtime when you run botty or during the env creation? If it is something like "module not found" I think our best bet would be to go with the heavy artillery, as they propose on their wiki to run this python Scripts/pywin32_postinstall.py -installhttps://github.com/mhammond/pywin32On 5 Jan 2022, at 20:34, aeon0 ***@***.***> wrote:That for example does not work for me on one machine... I am a bit lost with this one : /It seems different solutions work for different ppl but there is no one solution that works for all...—Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications on the go with GitHub Mobile for iOS or Android. You are receiving this because you authored the thread.Message ID: ***@***.***>

FrancescoZini:
running `python %CONDA_PREFIX%\Scripts\pywin32_postinstall.py -install` after `conda activate botty` fix the issue for me.
I've tried to revert to the pywin32==225 version of the code (which is broken for me) and run that line, after that it worked fine.
If that's fine we can either add that in the README.md or I can create an install script to run it after installation (I need to spend some time looking into it as I'm not practical of either Windows nor conda and seems not straightforward to add scripts to be run after the activate)

aeon0:
This is good to know. Lets wait how many ppl run into issues. I think very often it is also just a matter of not updating the conda env properly....

PR #395, opened 2022-01-05
utilize per-NPC ROI's and pose positions

mgleed:
Added NPC-specific ROI's to search for templates. Also included various NPC pose positions and then sorted template match results by composite of distance to nearest pose position and template match score. Also added a few more templates for a couple troublesome NPC's. Performance seems pretty good.

https://user-images.githubusercontent.com/9866239/148165005-fba544e1-96a2-4ba8-948e-8970a2791109.mp4

.


aeon0:
More tokio drift prs please!

PR #396, opened 2022-01-05
Test utils misc load_template()

aeon0:


aeon0:
Even with this change I see some ppl still getting that error... Not sure.

PR #398, opened 2022-01-05
correcting a4 pathing issue

skizotenigma:


aeon0:
Is this to move infront of the pillar for telekensis?

PR #401, opened 2022-01-05
Buy greater potions in nightmare

bjoerm:
If super potions are not available (e.g. nightmare), buy greater potions.


aeon0 (approved):
Nice

PR #403, opened 2022-01-06
Char selector after restart

aeon0:


src/ui/char_selector.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def select_char(self):
+        if self._char_template is not None:
+            template_result = self._template_finder.search(
+                self._char_template,
+                self._screen.grab(),
+                threshold=0.8,
+                roi=self._config.ui_roi["char_selection_all"], normalize_monitor=True
+            )
+            if template_result.valid:
```
FrancescoZini, line 30: maybe here would be wise to scroll down if the character is not visible upon restart, so to have something like

```
scrolls=0
template_result = # init to non valid somehow
while not template_result.valid and scrolls < 3:
  scrolls+=1
  mouse.wheel(-14) # in my system 14 scrolls down basically almost refresh the characters and you would reach the max character within 3 char scrolls I believe
  template_result =...
```
that would also require a change in custom mouse of the following function:
```
    @staticmethod
    def wheel(delta):
        _mouse.wheel(delta)
```
To be fair I haven't tested yet if the delta of the mouse actually correspond to 1 tick of my mouse so that's an assumption for now.

aeon0: Good point. Will test tomrrow.

FrancescoZini: I've made the changes, please have a look, the only thing that is missing is identifying which character is currently selected rather than assuming the top character is the correct one

FrancescoZini: by the way I have the changes locally to identify the character with a 0.99 match I just need to clean up a bit the code, since it's late I'll probably do it tomorrow

FrancescoZini: Added the remaining changes, I added them as a pull request to this branch, rather than pushing directly to this branch as I did earlier. 
This way you can decide whether you want the changes or not

aeon0:
Many ppl report big issues with that one. Standing around with mule in a1 :D 
@FrancescoZini didnt get time to test yet. But will do shortly. If you also wanna take a look, would be great. Thx.

FrancescoZini:
yeah I think the problem with this one is the following:

the character you select to play botty with is not always the top one. So most likely the wrong template would be saved.

That is the case for me for example as the character I select is usually in position #16-17 in the char selection, that way even by scrolling down she won't be able to be at the top as there are not enough characters. The game pretty much never crashes for me so I don't really mind personally but that might be the issue people are coming across.

The change I proposed in the other pull request was to identify in the screen the golden rectangle (by just applying a color filter on the image) and then save the template of the character circled in gold.

Even my solution wouldn't work though if, for example, a person select a character (highlighted in the gold rectangle) then scroll up or down to bring the character out of vision and the start botty.

With my solution though we can have situations (the user scrolls the char out of vision) where we don't have a template saved (simply because the golden rectangle was not identified) in that case we should stop botty upon restart.


Sorry ti was a bit long, the long story short would be that the current approach is prone to "bug" (unless disabled by default and left only to use to people that knows how it works) because it makes a dangerous assumption: that the top character is the one that is going to play, rather than trying to identify which one is the one that is actually selected.

FrancescoZini:
by the way I still think this should be an optional feature, anyway I will re-raise againt he PR to save the currently selected char (golden rectangle filter) but I will also make a separate one to have this feature optional

FrancescoZini:
This is the PR to make the feature optional:
https://github.com/aeon0/botty/pull/417

And this is the PR that I raised a couple of days ago to save the template of the character currently selected by filtering on the golden rectangle:
https://github.com/aeon0/botty/pull/418

Life's a bit in the way today so I haven't re-tested these two functionalities properly yet, I'll probably do it this evening.

PR #405, opened 2022-01-07
Scroll the characters if the template was not found.

FrancescoZini:
*PULL REQUEST TO aeon0:char-selector NOT master*

Apologies if I initially committed my changes to your branch.

This is the remaining bit to identify which character is currently selected (provided its name appear in the right side) and save it.
Upon restart the game will now attempt to scroll down the character if the original character is not found.

If the original character cannot be found the bot stops


FrancescoZini:
I believe the checks are failing because they merge all the colors under game.ini, since I added two it's probably failing to find stuff, that shouldn't be a problem for the bot itself as it never merges all the color to find objects (or at least I couldn't find any occasion where it did).

Nonetheless I should probably move the two additional filter under another config that isn't 'colors'. Unfortunately I don't have time now as I'm about to go for a (minor) surgery. Will move the configs at some point either later or tomorrow.

aeon0:
lets merge the other pr first, then proceed with any improvements. But since it is most likely sth that is fixed in next path on d2r side, I wouldn't spend too much time on it and not willing to include any complicated logic for that which might make the source code less readable.

PR #408, opened 2022-01-07
Messaging update

skizotenigma:
Updates the messaging to add more capabilities to the default message and seperates the status from it.


src/messenger.py
```diff
@@ -30,19 +30,24 @@ def send_stash(self):
     def send_gold(self):
         self._message_api.send_gold()
 
-    def send_message(self, msg: str):
-        self._message_api.send_message(msg)
+    def send_status(self, msg: str):
+        self._message_api.send_status(msg)        
+
+    def send_message(self, title: str, msg: str, img: str):
+        self._message_api.send_message(title, msg, img)
```
aeon0, line 37: This call will throw an error for the generic API because it does not have img as an argument. You will have to add img as an optional argument e.g. with default = None

aeon0 (changes requested):

src/messenger.py
aeon0: You will have to add the "img" param to the send_message() function in generic_api.py

bjoerm:
Hey, there is a bug in this PR. send_message() gets in _send_status_update() only msg as input, but it also needs now title and optional img. This is currently stopping botty anytime it wants to send a message.

aeon0:
@skizotenigma I reverted the changes on master. Another problem that came up is that not all send_message() calls where updated to use the "title", that why you get empty messages with the generic_api.

Since I don't have access to your fork I can not "reopen" the PR. But feel free to do so.

PR #415, opened 2022-01-08
added refill chest key option

JohnStyleZ:


src/bot.py
```diff
@@ -273,10 +273,14 @@ def on_maintenance(self):
 
         # Check if we are out of tps or need repairing
         need_repair = self._ui_manager.repair_needed()
-        if self._tps_left < random.randint(3, 5) or need_repair or self._config.char["always_repair"]:
+        if self._tps_left < random.randint(3, 5) or need_repair or self._config.char["always_repair"] or self._config.char["always_refill_keys"] & self._config.routes["run_arcane"]:
```
aeon0, line 276: I mean it works. But it is a bit weird to use an bitwise and for bool if condition. I think it is more readable if you use  "and" here.

src/bot.py
```diff
@@ -273,10 +273,14 @@ def on_maintenance(self):
...
             if need_repair: Logger.info("Repair needed. Gear is about to break")
             else: Logger.info("Repairing and buying TPs at next Vendor")
-            self._curr_loc = self._town_manager.repair_and_fill_tps(self._curr_loc)
+            if self._config.char["always_refill_keys"] & self._config.routes["run_arcane"]:
```
aeon0, line 279: keep code dry. you use twice the same condition (always_refill_keys and run_arcane). Store it in a variable similar to how it's done with need_repair. Something like `buy_keys`

src/bot.py
```diff
@@ -273,10 +273,14 @@ def on_maintenance(self):
...
+                self._curr_loc = self._town_manager.refill_keys(self._curr_loc)
+                if not self._curr_loc:
+                	return self.trigger_or_stop("end_game", failed=True)
```
aeon0, line 282: Is the indentation correct here?

src/bot.py
```diff
@@ -273,10 +273,14 @@ def on_maintenance(self):
...
+            else: self._curr_loc = self._town_manager.repair_and_fill_tps(self._curr_loc)
```
aeon0, line 283: Should this be in an else branch? Couldnt we need to refill keys AND might have to repair?

aeon0 (changes requested):
Whats the rule for vendors selling keys? All vendors that repair also sell keys? Shouldnt we have one function that basically does all things: "repairs", "restocks tps" and, if selected in the configs, also buys keys?

JohnStyleZ:
> Whats the rule for vendors selling keys? All vendors that repair also sell keys? Shouldnt we have one function that basically does all things: "repairs", "restocks tps" and, if selected in the configs, also buys keys?

the way it refill keys just like refilling tps, you have to have a key in your inventory in order for it to work, also since you are already at the vendor, i dont see why not just repair and refill tps too. the refill key only works if arcane and refill key are selected, cuz you dont need it for other runs.

JohnStyleZ:
> Whats the rule for vendors selling keys? All vendors that repair also sell keys? Shouldnt we have one function that basically does all things: "repairs", "restocks tps" and, if selected in the configs, also buys keys?

the way it refill keys just like refilling tps, you have to have a key in your inventory in order for it to work, also since you are already at the vendor, i dont see why not just repair and refill tps too. the refill key only works if arcane and refill key are selected, cuz you dont need it for other runs.

src/bot.py
JohnStyleZ: It refill key also repair and refill tps

src/bot.py
JohnStyleZ: fixed

src/bot.py
aeon0: Thats my problem, cant we reuse the already implemented repair_and_tp to take a flag `refill_keys: bool = False`. It looks like pretty much duplicated code now in the town manager.

aeon0 (changes requested):

src/bot.py
JohnStyleZ: since we are already at the vendor, why not repair and tp?

mgleed:
Thank you for your work on this. Given aeon's advise to refactor in order to not have to copy `repair_and_tp` and no recent updates, will close for now. Feature is coming soon, though. #484 will purchase keys, pickup keys on the ground if needed, and will decide if you want to use keys by checking for them in inventory

PR #417, opened 2022-01-09
D2r restart optional

FrancescoZini:
D2R restart feature can now be disabled via the config


aeon0:
The problem is not the d2r restarting. But the char selector

FrancescoZini:
True, but I don’t think we can implement a 100% reliable way to correctly select the right character without an OCR to parse the character name and recognize it upon restart. That’s why I think this feature should be configuarble for the time being.

aeon0:
> True, but I don’t think we can implement a 100% reliable way to correctly select the right character without an OCR to parse the character name and recognize it upon restart. That’s why I think this feature should be configuarble for the time being.
> […](#)
> Sent from my iPhone
> On 9 Jan 2022, at 15:11, aeon0 ***@***.***> wrote: ﻿ The problem is not the d2r restarting. But the char selector — Reply to this email directly, view it on GitHub, or unsubscribe. Triage notifications on the go with GitHub Mobile for iOS or Android. You are receiving this because you authored the thread.

My point is, that the feature switch should also turn on/off the char selection part

FrancescoZini:
Ahhh right right yeah I might have missed that but you are right. I kind of consider d2 restart and char selection the same feature in my head

aeon0 (changes requested):
- Also dont execute the char selection in game_controller.py if restart_d2r_when_stuck=0
- Make restart_d2r_when_stuck default = 0
- Add docu in readme.md and specify that if the feature is used the char must be the top most when starting botty! (Maybe also add this comment to the params.ini to make sure ppl see it)

FrancescoZini:
> * Also dont execute the char selection in game_controller.py if restart_d2r_when_stuck=0
> * Make restart_d2r_when_stuck default = 0
> * Add docu in readme.md and specify that if the feature is used the char must be the top most when starting botty! (Maybe also add this comment to the params.ini to make sure ppl see it)

Sure, will make all the changes as soon as I get some free time, hopefully tomorrow evening

FrancescoZini:
Done! Apologies for the delay...

src/game_controller.py
```diff
@@ -86,20 +87,25 @@ def run_bot(self, pick_corpse: bool = False):
...
+                        self.start_death_manager_thread()
+                        self.game_recovery = GameRecovery(self.screen, self.death_manager)
+                        return self.run_bot(True)
+                Logger.error(f"{self._config.general['name']} could not restart the game. Quitting.")
+                messenger.send_message("Got stuck and could not restart the game. Quitting.")
+            else:
+                Logger.error(
+                    f"{self._config.general['name']} could not recover from a max game length violation. Quitting botty.")
+                if self._config.general["custom_message_hook"]:
+                    messenger.send_message(f"{self._config.general['name']}: got stuck and will now quit botty")
```
aeon0, line 108: No need to add `{self._config.general['name']}`. The api implementations take care when to add the name and when not.
I know, in some places we still need to remove it. It currently will send the name twice.

src/game_controller.py
```diff
@@ -86,20 +87,25 @@ def run_bot(self, pick_corpse: bool = False):
...
+                Logger.error(
+                    f"{self._config.general['name']} could not recover from a max game length violation. Restarting the Game.")
+                if self._config.general["custom_message_hook"]:
+                    messenger.send_message(f"{self._config.general['name']}: got stuck and will now restart D2R")
+                if restart_game(self._config.general["d2r_path"]):
+                    self.game_stats.log_end_game(failed=max_game_length_reached)
+                    if self.setup_screen():
+                        self.start_health_manager_thread()
+                        self.start_death_manager_thread()
+                        self.game_recovery = GameRecovery(self.screen, self.death_manager)
```
aeon0, line 100: GameRecover() takes the self._template_finder as param now.

aeon0 (changes requested):
2 mini issues. Then we are good. Thx! This is an important fix for next release : )

src/game_controller.py
FrancescoZini: that's what happens when I do things in a rush :D sorry for that. It was also highlighted by my IDE

src/game_controller.py
```diff
@@ -86,20 +87,25 @@ def run_bot(self, pick_corpse: bool = False):
...
-                
+            if self._config.general['restart_d2r_when_stuck']:
+                Logger.error(
+                    f"{self._config.general['name']} could not recover from a max game length violation. Restarting the Game.")
+                if self._config.general["custom_message_hook"]:
+                    messenger.send_message(f"{self._config.general['name']}: got stuck and will now restart D2R")
+                if restart_game(self._config.general["d2r_path"]):
+                    self.game_stats.log_end_game(failed=max_game_length_reached)
+                    if self.setup_screen():
+                        self.start_health_manager_thread()
```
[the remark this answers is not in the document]
aeon0: I was thinking the same thing but didnt want to mess with it :D

aeon0 (approved):

PR #418, opened 2022-01-09
Char selector with scroll

FrancescoZini:
Char selector now identifies the character based on the golden rectangle around the one currently selected


aeon0 (changes requested):
You could just make the template transparent and would get the same results, no? Cause currently the color filtering is basically used to do exactly that: create a mask.

Apart from that, I had bad experience matching templates with just the outline. We tried this with the merc img once and for about 50% of ppl it didnt work out well.

Maybe its better to just filter by color and see how many pixels are left and compare that to others.

All in all I almost feel this char selection is causing more trouble than its worth : /

FrancescoZini:
> You could just make the template transparent and would get the same results, no? Cause currently the color filtering is basically used to do exactly that: create a mask.
> 
> Apart from that, I had bad experience matching templates with just the outline. We tried this with the merc img once and for about 50% of ppl it didnt work out well.
> 
> Maybe its better to just filter by color and see how many pixels are left and compare that to others.
> 
> All in all I almost feel this char selection is causing more trouble than its worth : /

Good point, I will try with transparency and see how it goes, so far this color_filter on both image and template gave me a pretty good match, between 0.98 to 1.0, but I haven't tried with more than 15 different characters so not sure how reliable it might be for everyone else.

I kind of suck at computer vision to be honest, all I know I learned it the past few days by reading your code lol (which is very nice by the way!)

aeon0:
I dont want to add more complexity to the automatic char selection. I think its good enough documented and default off which is enough to me. sry.

PR #420, opened 2022-01-10
Fix generic api messages

Master-Matt:
This is a quick fix to correct the generic api messages.
No need to do the full pr if you want to just add it in yourself.


aeon0:
Closing in favour of #422

PR #436, opened 2022-01-12
Inject Config instead of spawning new instances

hectorgimenez:
First step to move from instantiating a Config() each time to reuse the same instance.

I think we can investigate a bit if all the places where we are using config is really required or can be handled in a different way or config was there only to instantiate other dependencies that can also be injected.

Next steps could be stop using dictionary base access to the config parameters and use class attributes, this will help to reduce possible discrepancies like config options renaming breaking stuff, but not sure if you want to do it or you like to stay with the dictionary based access @aeon0 

Also, I see many tests failing but not sure why CI job is passing


FrancescoZini:
I wanted to do something similar as well, my concern was the fact that the config is being used concurrently by several threads, there are only few lines of code (read: that I could find, I didn't spend much time on this tbh) that actually modify the config and those do not happen often, still they do happen.

In addition, there are currently some workflow within botty that might spawn additional threads without killing the old ones. That is something to be addressed, that is true. I'm also not sure whether these threads do actively modify the config or not.

My 2 cents here would be to also make the Config class accessible in a thread safe way. That would probably require to have a dedicated class to handle the config, whether you have it attribute or dictionary based I don't think really mind, personally. @aeon0 might have a stronger opinion on this though.

That said, I just wanted to give my opinion on this, and I acknowledge also the fact that race-condition with the current implementation would probably be extremely rare to happen, if ever.

PR #445, opened 2022-01-13
Gambling

RB-Development:
Added gambling support.
"Advanced_Lootfilter" should be merged first.
This branch will detect if stash is full of gold. Afterwards the gambling functionality (currently only in act 4 implemented) will be called. The bot will gamble for specific items (determined in params.ini) until no gold is left. He will check for properties specified in pickit.ini and sell not matching items. 
Supported items are: ring, amulet, circlet, coronet, talon


aeon0:
It seems your other PR about advanced pickit is also included in here. Maybe we should first try to merge that one, then move here.

RB-Development:
Merged branch master into gambling and did some bugfixing. Ready for first review

config/params.ini
```diff
@@ -89,6 +89,11 @@ id_items=1
 open_chests=1
 fill_shared_stash_first=0
 
+[gamble]
+gamble=1,
```
mgleed, line 93: let's make it opt-in to gamble, so default `=0`, what do you think?

src/ui/ui_manager.py
```diff
@@ -602,6 +642,75 @@ def repair_needed(self) -> bool:
...
+    def gamble(self, item_finder: ItemFinder):
+        gold = True
+        gamble_on = True
+        if self._config.char["num_loot_columns"]%2==0:
+            ignore_columns = self._config.char["num_loot_columns"]-1
+        else:
+            ignore_columns = self._config.char["num_loot_columns"]-2
+        template_match = self._template_finder.search_and_wait(
+                "REFRESH",
+                best_match=True,
```
mgleed, line 661: `best_match` is only used if there multiple templates being searched for and you want to find the one with the closest match on screen.

src/ui/ui_manager.py
```diff
@@ -602,6 +642,75 @@ def repair_needed(self) -> bool:
...
+                        template_match = self._template_finder.search ("REFRESH", self._screen.grab())
+                        if (template_match.valid):
+                            x, y = self._screen.convert_screen_to_monitor(template_match.position)
+                            mouse.move(x, y, randomize=12, delay_factor=[1.0, 1.5])
+                            wait(0.1, 0.15)
+                            mouse.click(button="left")
+                            wait(0.1, 0.15)
+                        template_match_item = self._template_finder.search (item.upper(), self._screen.grab(), roi=self._config.ui_roi["vendor_stash"])
+                    #item found in gambling menu
+                    x, y = self._screen.convert_screen_to_monitor(template_match_item.position)
```
mgleed, line 684: anywhere you're using `convert_screen_to_monitor()` on a template match, you can simply add `normalize_monitor = True` to the arguments of `template_finder.search()` and then do `mouse.move(*template_match.position , ...` See #499

mgleed (changes requested):

config/params.ini
RB-Development: i agree. Default should ge gamble=0

src/ui/ui_manager.py
RB-Development: Good to know. Set best_match to False

config/params.ini
```diff
@@ -89,6 +89,11 @@ id_items=1
 open_chests=1
 fill_shared_stash_first=0
 
+[gamble]
```
mgleed, line 92: tbh I think we could even simplify futher. Get rid of `[gamble]` and `gamble=0` and just put `gamble_items=` under `[general]` or something. then update logic of the functions to instead of saying `if gamble:` do `if gamble_items:` That way we only add one line to config. If you don't want to gamble just leave the `gamble_items` blank. I know aeon's been trying to trim the config files based on his most recent commits.

mgleed (approved):

codahq:
@mgleed @RB-Development 
so...  as it stands it doesn't appear that you can turn off gambling.  the new setting "gamble_items=" can't be left.  if it is the bot will crashing during maintenance once full gold stash is reached because line 661 of ui_manager.py just blindly passes the first element of the gamble_items string without checking it for safety.  we need to do some checking so that if no gamble items are supplied it doesn't attempt to go gamble and just turns off gold pickup as it did before.  i don't want the bot to gamble for me.

codahq:
i've found another sort of annoying side effect of this commit.  the bot will gamble all of its personal tab, all of tabs 2 and 3 and most of the fourth tabs gold away.  i'm sure this is by design, however, the bot leaves no gold in reserve to resurrect a mercenary and doesn't know how to pull gold from the fourth tab for this purpose.  it probably shouldn't go to zero unless you teach it to pull gold from shared tabs to resurrect its mercenary.

PR #448, opened 2022-01-14
Nihlathaks no longer count cold immune run as fail

D2RLegit:
Changed false to True


aeon0:
Hmm, I understand that it is nicer to tp out instead of quitting the game... But counting it as a successful run will kind of screw with the stats.

PR #451, opened 2022-01-14
Make config static and global

aeon0:
@hectorgimenez please have a look

### DIFF
--- a/src/bot.py
+++ b/src/bot.py
@@ -46,7 +46,7 @@ class Bot:
         self._messenger = Messenger()
         self._config = Config()
         self._template_finder = template_finder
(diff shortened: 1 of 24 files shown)


FrancescoZini (approved):
Looks good to me. Again I don't have a strong opinion between dependency injection and global config, in this case I like this approach more as I believe this better separates the bot from the config.

hectorgimenez (approved):
Yup, I usually prefer DI over the global config but because usually all parts can be initialized with the correct configuration and then not directly accessed anymore. But this application is small compared to the amount of parameters and the different places where they are accessed, so I think in this case this approach is cleaner. LGTM

PR #453, opened 2022-01-15
Find D2R window using win32api instead of pattern matching

VladimirMakaev:
### Why
Currently to initialize Screen object we need to start with character selection page (or hack it). This is problematic when I want to have a faster feedback loop on testing changes. E.g. not restarting the game when testing out some pathing tweaks. Having Screen object being able to initialize any time we can write tests that navigate the bot mid game and just execute them via 
```
python test/test_name.py
```
Those tests are quite easy to do and I was able to repurpose smoke_test.py to use a real screen instead of mock screen and test scenario over and over without closing the game
### How
Win32 API allows cycling through the opened window and get the process that running them. This in turn allows getting ClientRect of the window without window border. Which is pretty much what old code did. 

### Test Plan
1. I tested that the offsets returned by both methods are exactly the same.
2. I've also put this new code under config setting having it off by default in case it's unstable
3. I haven't tested multi-monitor setup as I have one large monitor

### DIFF
--- a/src/screen.py
+++ b/src/screen.py
@@ -5,6 +5,7 @@ import time
 from logger import Logger
 from typing import Tuple
 from config import Config
+from utils import misc
 from utils.misc import load_template
 import os
 
@@ -24,12 +25,42 @@ class Screen:
         self._config = Config()
         self._monitor_roi = self._sct.monitors[monitor_idx]
         # auto find offests
+        self.found_offsets = False        
+        position = None
+        if self._config.general["find_window_via_win32_api"] :
+            Logger.debug("Using WinAPI to search for window under D2R.exe process")    
+            position = self.find_window_via_winapi()
+            if position is None:
+                Logger.debug("Can't find any window owned by D2R.exe falling back to matching via assets. Make sure D2R is in focus and you are on the hero selection screen")
+        
+        if position is None:                
+            position = self.find_window_via_assets(wait)        
+
+        if position is not None:
+            self._set_window_position(*position)
+        else:
+            if self._config.general["info_screenshots"]:
+                cv2.imwrite("./info_screenshots/error_d2r_window_not_found_" + time.strftime("%Y%m%d_%H%M%S") + ".png", self.grab())
+            Logger.error("Could not find hero selection or template for ingame, shutting down")
+            Logger.error("Could not determine window offset. Please make sure you have the D2R window " +
+                                    f"focused and that you are on the hero selection screen when pressing {self._config.general['resume_key']}")
+    
+    def _set_window_position(self, offset_x: int, offset_y: int):
+        Logger.debug(f"Set offsets: left {offset_x}px, top {offset_y}px")
+        self._monitor_roi["top"] += offset_y
+        self._monitor_roi["left"] += offset_x
+        self._monitor_x_range = (self._monitor_roi["left"] + 10, self._monitor_roi["left"] + self._monitor_roi["width"] - 10)
+        self._monitor_y_range = (self._monitor_roi["top"] + 10, self._monitor_roi["top"] + self._monitor_roi["height"] - 10)
+        self._monitor_roi["width"] = self._config.ui_pos["screen_width"]
+        self._monitor_roi["height"] = self._config.ui_pos["screen_height"]
+        self.found_offsets = True
+
+    
+    def find_window_via_assets(self, wait: int) -> Tuple[int, int]:
         template = load_template(f"assets/templates/main_menu_top_left.png", 1.0)
         template_ingame = load_template(f"assets/templates/window_ingame_offset_reference.png", 1.0)
-        start = time.time()
-        self.found_offsets = False
-        Logger.info("Searching for window offsets. Make sure D2R is in focus and you are on the hero selection screen")
         debug_max_val = 0
+        start = time.time()
         while time.time() - start < wait:
             img = self.grab()
             self._sct = mss()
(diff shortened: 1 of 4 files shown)


codahq:
What about people running botty in a vm and botty on the host? This should be configurable as an option or blocked entirely.

VladimirMakaev:
> What about people running d2 in a vm and botty on the host? This should be configurable as an option or blocked entirely.

This should work on VM and it’s also off by default via find_window_via_win32_api confif setting

codahq:
> This should work on VM

How?  Is it common for virtual machine architectures to passthrough the win32_api from the vm to the host?  I doubt that. 
 I have a feeling that botty running on the **host** and d2r running the **vm** can't work.

VladimirMakaev:
> > This should work on VM
> 
> How? Is it common for virtual machine architectures to passthrough the win32_api from the vm to the host? I doubt that. I have a feeling that botty running on the **host** and d2r running the **vm** will make win32_api calls to find the window impossible.

You’re right. I missed the fact you’re running the bot on the host. Still it’s disabled by default though so the old way will be used.

dboice:
Perhaps an win32api call to identify the window, if none is found, it falls back to the current implementation. Best practices: use the most reliable method first, and handle edge cases second. Just my two cents.

VladimirMakaev:
> Perhaps an win32api call to identify the window, if none is found, it falls back to the current implementation. Best practices: use the most reliable method first, and handle edge cases second. Just my two cents.

This exactly how it works if find_window_via_win32_api=1

src/screen.py
```diff
@@ -24,12 +26,36 @@ def __init__(self, monitor: int = 0, wait: int = 20):
...
-        start = time.time()
         self.found_offsets = False
         Logger.info("Searching for window offsets. Make sure D2R is in focus and you are on the hero selection screen")
+        position = self.find_window_via_winapi() if self._config.general["find_window_via_win32_api"] else None
+        if position is None:
+            position = self.find_window_via_assets(wait)
+        if position is not None:
+            self._set_window_position(*position)
+        else:
+            if self._config.general["info_screenshots"]:
```
FrancescoZini, line 37: I think I spent like 10 minutes reading these lines, I think it would be clearer to write something like this:

`
position = self.find_window_via_winapi() if self._config.general["find_window_via_win32_api"] else self.find_window_via_assets(wait)

if position is not None:
  ...
`

that should have the same outcome no?

FrancescoZini:
I'm ok with this change generally, the only slight concern is that this is moving botty a bit more towards memory reading rather than pure pixel bot. This should not be a problem in this case, I'm just considering that this is moving the bar a little bit. Win32API is not straightforward and many calls there can be intercepted via various hooks provided by the lib itself (hence increasing programmatic detectability, I'm frankly not sure about these ones tbh and I don't think this should be an issue considering we run this only a constant number of times rather than continuosly while botty runs). If @aeon0 is happy with this I'm happy as well

src/screen.py
VladimirMakaev: Not really. self.find_window_via_winapi() will return None in case it can't find the window using WinAPI. So in the code you're suggesting we won't have a fallback in case WinAPI fails. But let me do a change here to get this more readable and also fix misleading logging information

VladimirMakaev: I've pushed an update. It should be more readable now I think with log messages reflecting the process

src/screen.py
```diff
@@ -24,12 +26,38 @@ def __init__(self, monitor: int = 0, wait: int = 20):
...
         self._monitor_roi = self._sct.monitors[monitor_idx]
         # auto find offests
+        self.found_offsets = False        
+        if self._config.general["find_window_via_win32_api"] :
+            Logger.debug("Using WinAPI to search for window under D2R.exe process")    
+            position = self.find_window_via_winapi()
+            if position is None:
+                Logger.debug("Can't find any window owned by D2R.exe falling back to matching via assets. Make sure D2R is in focus and you are on the hero selection screen")
+                position = self.find_window_via_assets(wait)
+        if position is not None:
```
FrancescoZini, line 36: well now this position will not be initialize here though unless I'm missing something, have you tried running this code with the config disabled? is it fine?

VladimirMakaev: You're right let me get that fixed

VladimirMakaev: Tested 4 scenarios, fixed a bug with position not being initialized

FrancescoZini: thanks I find it clearer now

FrancescoZini (approved):

FrancescoZini:
hey,thanks for asking my review. I'm not sure how comfortable you are with me merging someone else changes, so I've only approved it for now.https://github.com/aeon0/botty/pull/453I'll leave the merge to you.On 17 Jan 2022, at 15:32, aeon0 ***@***.***> ***@***.*** requested your review on: #453 Find D2R window using win32api instead of pattern matching.—Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications on the go with GitHub Mobile for iOS or Android. You are receiving this because your review was requested.Message ID: ***@***.***>

VladimirMakaev:
> I'm ok with this change generally, the only slight concern is that this is moving botty a bit more towards memory reading rather than pure pixel bot. This should not be a problem in this case, I'm just considering that this is moving the bar a little bit. Win32API is not straightforward and many calls there can be intercepted via various hooks provided by the lib itself (hence increasing programmatic detectability, I'm frankly not sure about these ones tbh and I don't think this should be an issue considering we run this only a constant number of times rather than continuously while botty runs). If @aeon0 is happy with this I'm happy as well

Thanks for your time reviewing this PR. I'd like to just clarify one thing regarding this. This fix is not memory reading at all since botty isn't accessing the address space of D2R.exe in any way. The information being queried e.g. process names & opened windows is already available to all applications by Windows OS itself and is already queried by software like Process Manager / etc. It's more about botty leveraging features of the host OS better rather than shifting towards memory checking. But that's just my opinion on the matter

aeon0:
Thanks guys for fighting it out : ) gj @VladimirMakaev

mgleed:
@VladimirMakaev @aeon0 

We might want to consider refactoring this some or saying that being on CSS is now a strict requirement. 

Previously you didn't technically need to be on character select screen when launching botty, but now it'll detect with win32api and start the other code like health/death monitoring, etc. Most notably `Saving top-most char as template` occurs before even being on the CSS.

```
[2022-01-28 21:43:16,618] DEBUG      Using WinAPI to search for window under D2R.exe process
[2022-01-28 21:43:16,620] DEBUG      Set offsets: left 320px, top 150px
[2022-01-28 21:43:17,292] INFO       Start health monitoring
[2022-01-28 21:43:17,292] INFO       Start Death monitoring
[2022-01-28 21:43:17,293] INFO       Saving top-most char as template
[2022-01-28 21:43:17,450] INFO       Setting up Hammerdin
[2022-01-28 21:43:17,454] INFO       Doing runs: odict_keys(['run_shenk', 'run_nihlathak', 'run_trav', 'run_arcane', 'run_diablo'])
[2022-01-28 21:43:17,457] INFO       Starting game #1
[2022-01-28 21:43:17,458] DEBUG      Waiting for Template ['MAIN_MENU_TOP_LEFT', 'MAIN_MENU_TOP_LEFT_DARK']
[2022-01-28 21:43:18,531] DEBUG      Found Match: MAIN_MENU_TOP_LEFT (100.0% confidence)
```

VladimirMakaev:
It's interesting cause it looked like before using WinAPI it was only working from CSS because it was finding the window based on the assets that are present on that particular page. Now it can work beyond CSS. At least Screen object can be created from anywhere and you can setup tests that you can run for parts of your code. What needs to be fixed I guess is those particular features to be aware that CSS might not be up and behave correctly.

Ideally I would like to be able to start the bot from anywhere and for it to be able to run. But I don't think this broke something that worked before as CSS was a strict requirement based on the way Screen object was created

mgleed:
Yes, this is mostly true. I guess I meant that the bot itself couldn't
start without CSS open, but you could technically hit F11 whenever you
wanted and the bot would start itself whenever you got to the CSS because
it waited for the template to set offsets.

Agree with you otherwise. You're probably right just making CSS a clearer
strict requirement for hitting F11 should be sufficient.

VladimirMakaev:
@mgleed I've pushed this PR #494 this should completely disambiguate mentioned behavior

PR #458, opened 2022-01-18
fix duplicate bot name in status update message

flapjax:
When the generic_api is in use, the bot name appears twice at the start of each status update message, because the bot name is added at the start by both the GameStats `_send_status_update` method and the GenericApi `_send` method.

example of bug:
example

This commit changes the behavior of the _send_status_update method. If the generic_api is in use, it will avoid adding the bot name to the start of the message.

### DIFF
--- a/src/game_stats.py
+++ b/src/game_stats.py
@@ -155,7 +155,10 @@ class GameStats:
         return msg
 
     def _send_status_update(self):
-        msg = f"{self._config.general['name']}: Status Report\n{self._create_msg()}\nVersion: {__version__}"
+        if self._config.general["message_api_type"] == "generic_api":
+            msg = f"Status Report\n{self._create_msg()}\nVersion: {__version__}"
+        else:
+            msg = f"{self._config.general['name']}: Status Report\n{self._create_msg()}\nVersion: {__version__}"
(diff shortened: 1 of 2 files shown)


src/game_stats.py
```diff
@@ -155,7 +155,10 @@ def _create_msg(self):
         return msg
 
     def _send_status_update(self):
-        msg = f"{self._config.general['name']}: Status Report\n{self._create_msg()}\nVersion: {__version__}"
+        if self._config.general["message_api_type"] == "generic_api":
+            msg = f"Status Report\n{self._create_msg()}\nVersion: {__version__}"
+        else:
+            msg = f"{self._config.general['name']}: Status Report\n{self._create_msg()}\nVersion: {__version__}"
```
aeon0, line 161: I would rather remove the name here and add it to the DiscordEmbeds class instead of this if clause

aeon0 (changes requested):

src/game_stats.py
flapjax: no prob, will do.

flapjax:
Made the changes and re-tested both api modes today. Good to go from my end.

PR #459, opened 2022-01-18
A3 Pathing fix.

skizotenigma:
Should correct the issue of portals in town being clicked on when walking past them (ormus->WP, WP ->Ormus etc)


aeon0:
we should just use force move, that will make 100% sure it does not happen

aeon0:
But if you say it is regardless of the issue at hand a nicer pathing we can merge.

skizotenigma:
> But if you say it is regardless of the issue at hand a nicer pathing we can merge.

well all the testing that was done it didn;'t happen anymore, but force move would work as well... i just went for the quick road lol. plus i always felt weird the way it ran around for cain and stashing.

PR #462, opened 2022-01-18
Switch to online

freesoul:
- Sometimes after exiting a game I had a "failed to load characters" or "failed to authenticate", which put me in the offline tab, and the game started with the wrong offline character. 
- I added the blacklist for the blue spiked club, that I found for some reason picked in my stash.


aeon0:
Spiked clubs stay!
The other feature is already included if you turn on `restart_d2r_when_stuck`. Just make sure the selected char is the top-most when you start botty.

freesoul:
> Spiked clubs stay! The other feature is already included if you turn on `restart_d2r_when_stuck`. Just make sure the selected char is the top-most when you start botty.

Restarting the complete process of d2r when in most of cases is enough to just click again on the "online" tab, is overkill though.

aeon0:
Another issue is that it will not allow for running botty offline anymore. Lots of testing is done offline.

PR #464, opened 2022-01-19
Fix teleport asset. Accept teleport amulet.

kod21236:


aeon0:
You should replace the existing ones instead of creating new assets since they are not used anywhere.

VladimirMakaev:
Doing similar things in this PR https://github.com/aeon0/botty/pull/476 but more.

PR #466, opened 2022-01-19
Added a class which can assist the quantification of item attributes

freesoul:
A class which assists the reading of numbers in order to assist the quantification of item attributes.
It is to be integrated later, together with the advanced pickit, so we will be able to include in the nomenclature something like "and(with_life[>18])".
It can be extended to read characters and other colors as well.


freesoul:
To add some context, I decided to try this dictionary based OCR as I trained models on custom datasets + the D2 fonts, and I got a lot of mistakes in the predictions. IDK if someone can get a model better good accuracy with Tesseract. It would be a pity to miss a 45hp skiller because it reads 15 istead of 45 ^^

mgleed:
Thank you for your work on this.

While #484 itself will likely not be mergeable, asset matching like this will become obsolete shortly. Will be splitting #484 into parts including the ocr library and features into a separate PR shortly. I got very good accuracy for these. You can test out the branch yourself if you'd like--it currently parses quantity from TP tomes, ID tomes, and keys and manages the counts. Will extend to other applications over time like item pickup, item filtering, etc.

PR #471, opened 2022-01-21
Fixed Vigor bug w/ walking Hammerdin in Trav

pokzcodes:
Very frequently when running Trav, a Hammerdin without teleport would get stuck using Vigor while casting hammers, significantly increasing run time. This change fixes that issue.

### DIFF
--- a/src/char/hammerdin.py
+++ b/src/char/hammerdin.py
@@ -114,7 +114,7 @@ class Hammerdin(IChar):
         # Check out the node screenshot in assets/templates/trav/nodes to see where each node is at
         atk_len = self._char_config["atk_len_trav"]
         # Go inside and hammer a bit
-        self._pather.traverse_nodes([228, 229], self, time_out=2.5, force_tp=True, do_pre_move=False)
+        self._pather.traverse_nodes([228, 229], self, time_out=2.5, force_tp=True)
         self._cast_hammers(atk_len)
         # Move a bit back and another round
         self._move_and_attack((40, 20), atk_len)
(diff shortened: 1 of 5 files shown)


src/char/hammerdin.py
```diff
@@ -114,7 +114,7 @@ def kill_council(self) -> bool:
         # Check out the node screenshot in assets/templates/trav/nodes to see where each node is at
         atk_len = self._char_config["atk_len_trav"]
         # Go inside and hammer a bit
-        self._pather.traverse_nodes([228, 229], self, time_out=2.5, force_tp=True)
+        self._pather.traverse_nodes([228, 229], self, time_out=2.5, force_tp=True, do_pre_move=False)
```
aeon0, line 117: Did you test with a teleporting pala that we are on teleport when this is called. It should, but I'd rather know it was tested.

aeon0 (changes requested):
I would merge the change if it wouldnt include all the changes from the other PR about the shared stash filling first.

src/char/hammerdin.py
pokzcodes: > 

Nope, only tested it walking. I will have to make a hero edited character to test

pokzcodes: > Did you test with a teleporting pala that we are on teleport when this is called. It should, but I'd rather know it was tested.

I am creating a new pull request that addresses this without affecting Enigma Hammerdins. Closing this one because my newer change is in a diff fork.

PR #472, opened 2022-01-21
Add option to fill shared stash tabs first (useful for HC characters)

pokzcodes:
Added an option fill_shared_stash_first to params.ini

When true, it will stash loot starting from the rightmost stash tab going left (personal tab last). That way dead HC characters can still at least recover the loot they found during their last session.


aeon0:
Nice, gj

PR #473, opened 2022-01-22
Fixed issue with vigor for walking Hammerdin in Trav (without breaking Enigma pal)

pokzcodes:
Very frequently when running Trav, a Hammerdin without teleport would get stuck using Vigor while casting hammers, significantly increasing run time. This change fixes that issue.

PR #474, opened 2022-01-23
Hammerdin-update-new

cutebeast:
Update to trav tele and walking pather to better clear trav. Also implemented the update on pindal based on information from Alternating#1115 & https://github.com/aeon0/botty/pull/473. Tested on both Tele and Walking Hammerdin and solving the issue from pathing instead.

### DIFF
--- a/src/char/hammerdin.py
+++ b/src/char/hammerdin.py
@@ -71,6 +71,12 @@ class Hammerdin(IChar):
         wait(0.1, 0.15)
         if self.can_teleport():
             self._pather.traverse_nodes_fixed("pindle_end", self)
+            # move back recast hammer
+            self._cast_hammers(self._char_config["atk_len_pindle"])
+            wait(0.1, 0.15)
+            self._move_and_attack((-350, 150), self._char_config["atk_len_pindle"]/2)
+            wait(0.1, 0.15)
+            self._pather.traverse_nodes_fixed("pindle_end", self)
         else:
             if not self._do_pre_move:
                 keyboard.send(self._skill_hotkeys["concentration"])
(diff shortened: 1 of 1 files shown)


jobithu:
Did pressure test this PR 75 runs vs 0.6.4

This PR:
1:40 avg by run & 
32 items over 75 runs
0 fails
0 MD
(2h 05min runtime)
0.25 items/min 

versus Botty 0.6.4: 
1:21 avg by run & 
20 items over 75 runs
0 fails
0 MD
(1h 40 runtime)
0.24 items/min

Whilst items/min is almost useless parameter for these durations tested, it seems a good idea to merge this PR to master with regard to TRAV.

Didn't test pindle though- not a fan of hammerdin pindle ;)

SwePow (approved):
Reviewed code and did 50 test runs. Merc seems to die more often but my Merc has Insight, Tal Mask and Treachery so with optimal gear this will happen less. Or with decent Pala in Trav Merc might not be needed at all.

src/char/hammerdin.py
```diff
@@ -113,19 +119,30 @@ def kill_council(self) -> bool:
             wait(0.05, 0.15)
         # Check out the node screenshot in assets/templates/trav/nodes to see where each node is at
         atk_len = self._char_config["atk_len_trav"]
-        # Go inside and hammer a bit
-        self._pather.traverse_nodes([228, 229], self, time_out=2.5, force_tp=True)
-        self._cast_hammers(atk_len)
         # Move a bit back and another round
         self._move_and_attack((40, 20), atk_len)
```
aeon0, line 123: We sure this should stay? It will always start with casting hammers outside. Also when teleport is avaiable.

src/char/hammerdin.py
```diff
@@ -113,19 +119,30 @@ def kill_council(self) -> bool:
...
+            self._move_and_attack((40, 20), atk_len)
             # Back to center stairs and more hammers
             self._pather.traverse_nodes([226], self, time_out=2.5, force_tp=True)
             self._cast_hammers(atk_len)
             # move a bit to the top
             self._move_and_attack((65, -30), atk_len)
+            self._move_and_attack((-125, -373), atk_len)
         else:
+            # Go to center stairs a bit
+            self._pather.traverse_nodes([226], self, time_out=2.5, force_tp=True, force_move=True)
```
aeon0, line 138: Isnt that the node we are at anyway after the traverse to "safe_dist"?

aeon0 (changes requested):

src/char/hammerdin.py
cutebeast: # Check out the node screenshot in assets/templates/trav/nodes to see where each node is at
        atk_len = self._char_config["atk_len_trav"]
        # Go inside and hammer a bit
        self._pather.traverse_nodes([228, 229], self, time_out=2.5, force_tp=True)
        self._cast_hammers(atk_len)
        # Move a bit back and another round
        self._move_and_attack((40, 20), atk_len)
        
        This is the original code for teleport

src/char/hammerdin.py
cutebeast: This will start hammer on the outside for this node before go in

D2RLegit:
I personally ran and reviewed this and do not like either of the scripts. hammerdin could use a couple small tweaks but this script is not it. The PIndle script is pointless. I would heavily advise on not merging pindle and having the attack sequence cut down on trav theres a lot of excess sequences that are not needed and poorly positioned. A good tweak will require offsetting nodes. I would not merge either of these.

mgleed:
I'm not very persuaded by the comments that the changes are helpful. Please make any needed adjustments and provide some comparison pre-vs-post run statistics (time spent per run, chicken percent, death percent, run fail percent, etc.) for both TP hammerdin and walking hammerdin. Closing for now while you work on those changes and provide the data--let me know when you've done so and we can reopen.

PR #475, opened 2022-01-23
Added a fix for walking characters getting stuck looting in Trav

pokzcodes:
Frequently, when a walking character ends up downstairs below the window in Trav while looting, and there are things inside the building, it will get stuck. To fix that, I added a step for characters that can't teleport that will walk them back through the door and inside the room to try looting again.


pokzcodes:
Closing this, I have improved it a lot more, will make another PR with better approach

PR #476, opened 2022-01-23
Optimized handling of teleport charges

VladimirMakaev:
## Why
Not everybody has Enigma and using amulet efficiently can be very helpful in Trav when you can guarantee entering the building with spending 2 charges then you can loot things downstairs and teleport back through the wall to pick up loot upstrairs. It makes trav runs on walkadin more reliable and fast
## How
The feature is behind *have_teleport_charges* config property and if you want to avail need to set it to 1

Since charges are an additional resource that we have it requires a explicit decision when using a charge is applicable. E.g. we do when traversing path nodes
E.g.
```
self._pather.traverse_nodes([228, 229], self._char, time_out=2.5, use_tp_charge=True)
```
Currently I'm only handling charges in trav on paladin but I've done the ground work to integrate it in other runs.

We also check charges by comparing assets with 1,2,3 charge on teleport and when it's 3 or less we repair. (to avoid doing always_repair)

Upon death and exit the keybind for teleport is lost. Therefore we do additional handling on corpse pickup and update the binding again

## Test Plan
1. Walkadin with have_teleport_charges=0 and teleport keybind set was teleporting the whole way succesfully consuming charges
2. Walkadin with have_teleport_charges=0 and teleport keybind not set was walking all the way without trying to teleport
3. Walkadin with have_teleport_charges=1 and teleport keybind set was walking up to the stairs teleported inside, cleared the room, teleport back to stairs and back after looting


jobithu:
I would recommend to rename "tp" to "tele" to avoid confusion.

TP = town portal 
Tele = teleport

VladimirMakaev:
> I would recommend to rename "tp" to "tele" to avoid confusion.
> 
> TP = town portal Tele = teleport
> 
> I would be interested in that PR for chaos- river of flame needs to be crossed using a Tele amu, but from there one can walk.

Actually pather already has pre-existing parameter "force_tp", I'm not really sure if it makes sense to rebrand throughout the codebase right now. Not in this PR probably

mgleed:
> > I would recommend to rename "tp" to "tele" to avoid confusion.
> > TP = town portal Tele = teleport
> > I would be interested in that PR for chaos- river of flame needs to be crossed using a Tele amu, but from there one can walk.
> 
> Actually pather already has pre-existing parameter "force_tp", I'm not really sure if it makes sense to rebrand throughout the codebase right now. Not in this PR probably

Agree that `tp` should be eliminated and changed (maybe `tele` and `town_portal`) from codebase including config params but yes different PR.

config/params.ini
```diff
@@ -97,6 +97,7 @@ always_repair=0
 id_items=0
 open_chests=1
 fill_shared_stash_first=0
+have_teleport_charges=0
```
mgleed, line 100: are you able to let the bot figure out whether teleport is a charged skill or not on 1st run? might be able to avoid having to add a param if so.

VladimirMakaev: This actually a good idea, less settings less questions why it doesn't work :)

VladimirMakaev: Done!

VladimirMakaev:
Implemented automatic detection of "capabilities" which can be extended later. But now it contain properties:
can_teleport_natively
can_teleport_with_charges
Can add something else to this later.

Just need to make sure we check for self.capabilities.<property> for those things and make sure we do so **AFTER** capabilities are discovered

mgleed:
> Implemented automatic detection of "capabilities" which can be extended later. But now it contain properties: can_teleport_natively can_teleport_with_charges Can add something else to this later.
> 
> Just need to make sure we check for self.capabilities. for those things and make sure we do so **AFTER** capabilities are discovered
> 
> **Note**: rebased on top of master and force-pushed. Staying on the bleeding edge :)

Changes look good to me, but I don't have a way to test this. You've tested in game? Concern I'd have would be template matching of charges 1-3, but I guess even if there is some false-positive matching it'd still trigger a need to recharge so maybe not a big deal.

VladimirMakaev:
I've tested walked in with / without charges and sorc with normal teleport. If somebody can test other classes with Enigma / without Enigma that would be great. At least not to break something that worked.

mgleed (approved):

PR #479, opened 2022-01-24
Improved walking Hammerdin in Trav (clears it more consistently w/o getting stuck and loots everything)

pokzcodes:
The current scripting for Hammerdins without Enigma results in a large variance in how much of the council are killed, and it very frequently gets stuck looting below or above the window.

One big problem with the current script is that it would start by running all the way inside the building. Whenever the council members spawned outside, your merc would tank them all outside while you sit inside casting Hammers at nothing. These changes now begin Hammers outside, at the entrance, so that is no longer an issue. The mobs will generally come out to meet you. Afterwards, it goes into the building to clean up anyone who stayed inside.

The other big problem with the script is with looting. The wall separating the interior of the building and the outside patio area caused the bot to get stuck looting on the wall - frequently skipping half or even all of the loot. My changes now check both the inside and outside for loot in separate steps when Teleport is not available.

I made sure to split these changes out in such a way that only a non-Teleporting character will be affected.

### DIFF
--- a/src/char/hammerdin.py
+++ b/src/char/hammerdin.py
@@ -113,22 +113,32 @@ class Hammerdin(IChar):
             wait(0.05, 0.15)
         # Check out the node screenshot in assets/templates/trav/nodes to see where each node is at
         atk_len = self._char_config["atk_len_trav"]
-        # Go inside and hammer a bit
-        self._pather.traverse_nodes([228, 229], self, time_out=2.5, force_tp=True)
-        self._cast_hammers(atk_len)
-        # Move a bit back and another round
-        self._move_and_attack((40, 20), atk_len)
         # Here we have two different attack sequences depending if tele is available or not
         if self.can_teleport():
+            # Go inside and hammer a bit
+            self._pather.traverse_nodes([228, 229], self, time_out=2.5, force_tp=True)
+            self._cast_hammers(atk_len)
+            # Move a bit back and another round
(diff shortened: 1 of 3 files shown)


Steve6030:
Often getting stuck outside trav. Specifically at this step [2022-01-24 18:28:47,421] DEBUG      Traverse: [230]

image

Steve6030:
This PR has been working great for my under geared walking hammerdin. 

Avg run 1:46 trav only. 

Very thorough drops and consistently walking through the door and killing council.

Tonst3r:
*Huge improvement from default walk.

The walking version will never be a perfect science, but this handles all the problems with blocking and looting very well. 100% this should be part of the master/default. Instead of derp-ing on a block, it continues trying to fight/move and even loot to be safe. Very quick runs too. 

Basic gear, no enigma, hoto/shako/hoz/cta but it's pretty-much overkill. 
70-80 second runs seem to be the avg. (My last 20 had 74 secs avg)

Great work, ty!

pokzcodes:
@aeon0 This PR should be good to go! I have a bunch of people currently running it. Here's a screenshot from today using these changes (note: I have a pretty selective pickit, so the number of items would be higher for others in the same number of runs):

image

pokzcodes:
Apparently, my changes did affect Enigma hammerdins, so I'm gonna fix that and re-submit

PR #480, opened 2022-01-25
Fixed CTA order - battle command needs to be before battle orders. Added several new assets for white/gray bows, swords and grandfather.

ggodev:


pokzcodes:
This seems like a good change, but it looks like you included some other changes/files that are not related to this one. I would bet that aeon will want you to split those changes out from the CTA change.

ggodev:
> This seems like a good change, but it looks like you included some other changes/files that are not related to this one. I would bet that aeon will want you to split those changes out from the CTA change.

Neither of the changes changes the already existing logic, but I could split them up if needed.

VladimirMakaev:
I think swapping keybinds around doesn't make much sense. Instead I'd suggest to make sure that bot is doing the right sequence. And think it does. E.g. It get's Battle Command first to get +1 and then Battle Orders to benefit from Battle Command that he casted before. 

So I'd remove CTA portion from here and rename PR to reflect the other changes. That's my 2 cents

ggodev:
> I think swapping keybinds around doesn't make much sense. Instead I'd suggest to make sure that bot is doing the right sequence. And think it does. E.g. It get's Battle Command first to get +1 and then Battle Orders to benefit from Battle Command that he casted before.
> 
> So I'd remove CTA portion from here and rename PR to reflect the other changes. That's my 2 cents

Honestly If the logic is already fixed (which it is), it's best to have the keybindings following that. 
It's a QoL change indeed, but it needs to be there imo.

mgleed (changes requested):
Happy to merge the item assets if you'd rename the PR. The other change involving the order of BC/BO in the params.ini file is not necessary. Besides, by updating the param.ini file you then reset the defaults of all users who have these keybindings and have not overwritten them in a custom.ini file which may lead to undesired downstream effects.

config/pickit.ini
```diff
@@ -1106,6 +1132,7 @@ uniq_weapon_ribcracker=1, ethereal
 uniq_hand_of_blessed_light=1,, ethereal
 uniq_heavens_light=1,, ethereal
 uniq_rune_master=1, ethereal
+uniq_weapon_grandfather=1
```
mgleed, line 1136: please make all of the new items you added have the first field be `=0` by default, users can then change them to 1 if desired.

mgleed:
No reply for ~48 hours, closing for now. Will reopen depending on interest in revising.

PR #481, opened 2022-01-25
Blizz sorc Chaos updates

D2RLegit:
Game ini pathing changes
diablo.py pickup item optimization
Blizz_sorc attack updates
Return to town for pots/stash at penta if needed.


D2RLegit (commented):
reviewed

PR #483, opened 2022-01-26
Support multiple monitors with win32api

AlekseyLesnoy:
- `find_window_via_win32_api=1` option only works with a single monitor setup, changes enables supporting multiple monitors
- Cleaning up some trailing whitespaces in screen.py


VladimirMakaev (approved):
lgtm

PR #485, opened 2022-01-27
prioritized looting

codahq:
the reason for this pr is because i watched the bot skip desirable items when there is a lot of gold drops with the desirable items like the end squares in arcane sanctuary where gold piles already sit on game creation.  currently looting fails because it hits the timeout when there are lots of gold piles potentially before good items are picked.  to fix this i made three attempts to loot; the first skips gold, the second skips potions, the third skips potions and gold.  the reason for this is to lessen the chances of leaving desirable loot behind that clipped off the screen.

it doesn't hurt anything and in anecdotal tests when i dropped a rune and enough gold to push the rune's label off the screen this method successfully achieved picking it up

### DIFF
--- a/src/item/pickit.py
+++ b/src/item/pickit.py
@@ -23,12 +23,12 @@ class PickIt:
 
     def pick_up_items(self, char: IChar, is_at_trav: bool = False) -> bool:
         """
-        Pick up all items with specified char
+        Pick up just items with specified char
         :param char: The character used to pick up the item
         :param is_at_trav: Dirty hack to reduce gold pickup only to trav area, should be removed once we can determine the amount of gold reliably
         :return: Bool if any items were picked up or not. (Does not account for picking up scrolls and pots)
         """
-        found_nothing = 0
+        
         found_items = False
         keyboard.send(self._config.char["show_items"])
         time.sleep(1.0) # sleep needed here to give d2r time to display items on screen on keypress
@@ -37,33 +37,66 @@ class PickIt:
             img = self._screen.grab()
             cv2.imwrite("./loot_screenshots/info_debug_drop_" + time.strftime("%Y%m%d_%H%M%S") + ".png", img)
             Logger.debug("Took a screenshot of current loot")
-        start = prev_cast_start = time.time()
-        time_out = False
+        
         picked_up_items = []
         skip_items = []
+        
+        Logger.debug("Looting all but gold")
+        found_items = self.pickup_by_type(char, True, False, picked_up_items, skip_items, is_at_trav, 10)
+        Logger.debug("Looting all but potions")
+        found_items = self.pickup_by_type(char, False, True, picked_up_items, skip_items, is_at_trav, 20) or found_items
+        Logger.debug("Looting items only (no gold and no potions)")
+        found_items = self.pickup_by_type(char, False, False, picked_up_items, skip_items, is_at_trav, 10) or found_items
+
+        keyboard.send(self._config.char["show_items"])
+        return found_items
+
+    def pickup_by_type(self, char: IChar, do_pots, do_gold, picked_up_items, skip_items, is_at_trav, time_out_time: int = 28) -> bool:
+        """
+        Pick up items with the ability to skip certain types of itesms
+        :param char: The character used to pick up the item
+        :param do_pots: Bool to determine if potions will be picked up
+        :param do_gold: Bool to determine if gold will be picked up
+        :param picked_up_items: The container of picked up items
+        :param skip_items: The container of skipped up items
+        :param is_at_trav: Dirty hack to reduce gold pickup only to trav area, should be removed once we can determine the amount of gold reliably
+        :param number of seconds to try to pickup the list
+        :return: Bool if any items were picked up or not. (Does not account for picking up scrolls and pots)
+        """
+        found_nothing = 0
+        found_items = False
+        start = prev_cast_start = time.time()
+        time_out = False
         curr_item_to_pick: Item = None
         same_item_timer = None
         did_force_move = False
         while not time_out:
-            if (time.time() - start) > 28:
+            if (time.time() - start) > time_out_time:
                 time_out = True
-                Logger.warning("Got stuck during pickit, skipping it this time...")
+                Logger.warning("Got stuck during pickit, skip looting with pots {} and gold {}...".format(do_pots, do_gold))
                 break
             img = self._screen.grab()
             item_list = self._item_finder.search(img)
 
-            # Check if we need to pick up certain pots more pots
-            need_pots = self._belt_manager.get_pot_needs()
-            if need_pots["mana"] <= 0:
+            if do_pots:
+                # Check if we need to pick up certain pots more pots
+                need_pots = self._belt_manager.get_pot_needs()
+                if need_pots["mana"] <= 0:
+                    item_list = [x for x in item_list if "mana_potion" not in x.name]
+                if need_pots["health"] <= 0:
+                    item_list = [x for x in item_list if "healing_potion" not in x.name]
+                if need_pots["rejuv"] <= 0:
+                    item_list = [x for x in item_list if "rejuvenation_potion" not in x.name]
+            else:
                 item_list = [x for x in item_list if "mana_potion" not in x.name]
-            if need_pots["health"] <= 0:
                 item_list = [x for x in item_list if "healing_potion" not in x.name]
-            if need_pots["rejuv"] <= 0:
                 item_list = [x for x in item_list if "rejuvenation_potion" not in x.name]
 
             # TODO: Hacky solution for trav only gold pickup, hope we can soon read gold ammount and filter by that...
-            if self._config.char["gold_trav_only"] and not is_at_trav:
+            if (self._config.char["gold_trav_only"] and not is_at_trav) or not do_gold:
                 item_list = [x for x in item_list if "misc_gold" not in x.name]
+            
+            #TODO always do items i guess or give an option to filter items for some reason?
 
             if len(item_list) == 0:
                 # if twice no item was found, break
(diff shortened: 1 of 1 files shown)


mgleed:
Could you benchmark this for us vs. 0.6.4?

I could see calling `item_finder.py` multiple times to be time-consuming. I think you could accomplish this same thing but more efficiently by implementing this logic in an inner loop... perhaps modifying the logic of the else statement in pickit.py

```py
            if len(item_list) == 0:
                # if twice no item was found, break
                found_nothing += 1
                if found_nothing > 1:
                    break
                else:
                    # Maybe we need to move cursor to another position to avoid highlighting items
                    pos_m = self._screen.convert_abs_to_monitor((0, 0))
                    mouse.move(*pos_m, randomize=[90, 160])
                    time.sleep(0.2)
            else:
                found_nothing = 0
                closest_item = item_list[0]
                for item in item_list[1:]:
                    if closest_item.dist > item.dist:
                        closest_item = item
```

codahq:
thank you for the reply and comment.  it is important to understand that if enough loot piles exist on the ground item labels are clipped.  not all loot piles on screen will have labels.  the whole point of this change **is to call item_finder again** after taking a new screenshot after some loot has already been picked up.  the point of the change in behavior is to clear some of the gold so that new labels of items on the ground for potentially higher value loot have room to appear.  

the change would also makes parameterizing the timeout for looting gold and potions trivial.  

this is applicable anywhere but where it is especially applicable is on the platform at the end of arcane sanctuary.  gold piles are generated on the ground during map generation.  in addition, there are monster drops.  and on top of that there are chest drops.  this is a very high potential for item labels to clip off screen.  i watched the bot leave behind a high rune because it didn't see it initially.  it was uncovered only after picking up some piles of gold and then it didn't prioritize it because it was far away.  the only way for it to know about an item that is coming onto the screen later is to take a new screenshot and call item_finder again, correct?

as far as benchmarking goes, it makes the finding items part of looting exactly three times more expensive because it calls item_finder three times.  it increases from N time to find items to 3N time to find items.  It still takes exactly the same amount of time per item but it may on occasion find additional items that it couldn't have found before.  on my system, the call to item_finder takes just under 1 second so it adds just under 2 seconds per call to pick_up_items.  this is very worthwhile to me because the entire point of the bot is to pickup valuable items.  there is absolutely no prioritization right now and this, at the very least, is a stop gap.  

if this blocks merging i would agree to calling item_finder just twice; once for gold/potions/items and then once more for just items (no gold, no potions).

mgleed:
> the whole point of this change **is to call item_finder again** after taking a new screenshot after some loot has already been picked up. the point of the change in behavior is to clear some of the gold so that new labels of items on the ground for potentially higher value loot have room to appear.
If the point of the update is to clear gold first to make room for more valuable item labels, then why do you prioritize gold last?

> i watched the bot leave behind a high rune because it didn't see it initially. it was uncovered only after picking up some piles of gold. even though it was on the screen after looting began it wasn't initially so the bot never knew about it. 
A new screenshot is captured and item_finder.search() is repeated already for every item that is picked up in vanilla. It generates a new `item_list` with a new screenshot until no items with any template matches remain on screen. If a monster dies while you are in the middle of picking items, you will see the new drops. Even if no items are found in the first pass, it tries again just in case.
```py
        while not time_out:
            if (time.time() - start) > 28:
                time_out = True
                Logger.warning("Got stuck during pickit, skipping it this time...")
                break
            img = self._screen.grab()
            item_list = self._item_finder.search(img)
```

> the only way for it to know about an item that is coming onto the screen later is to take a new screenshot and call item_finder again, correct? 
Yep, this is exactly how pickit already works as discussed above. 

> as far as benchmarking goes, it makes the finding items part of looting exactly three times more expensive because it calls item_finder three times. 
This is too expensive. Pickit processing already takes a long time, and affecting *every* use of pickit throughout the entire bot by tripling the calls to item_finder would be too much to ask. Part of the difficulty with this is that there are several areas where pickit is called while you are still potentially being hit by monsters and would greatly increase the risk of death/chicken.

>there is absolutely no prioritization right now and this, at the very least, is a stop gap.
I don't see how prioritizing items in the fashion you have solves the problem you describe. Let's say you have a HR drop, but that HR doesn't have an item label because there are too many stacks of gold/pots on the ground. Now you run pickit. Item_finder finds THE SAME items regardless of what you prioritize. So if you prioritize gold first, it'll pickup all the gold and hopefully the HR will come into view. But if you prioritize HR first, then pots, then gold; you will skip HR, pick pots, and still pick the same gold. Then botty, as it already does before this PR, will take a new screenshot and see the HR. You end up picking the exact same items.

codahq:
> If the point of the update is to clear gold first to make room for more valuable item labels, then why do you prioritize items before gold? `Logger.debug("Looting all but gold")` is first.

this one is easy.  the bot does chicken frequently and if a high rune is visible immediately i want the bot to pick it up before it ends up having to chicken.  in addition, sometimes the potions are waning by the end of the fight so i allow the bot to pick those up immediately as well so they can be available if necessary while it is being popped in the head by the remaining monsters while it is trying to grab what's important.  this is more well thought out than you realize.

> A new screenshot is captured and item_finder.search() is repeated already for every item that is picked up in vanilla. It generates a new `item_list` with a new screenshot until no items with any template matches remain on screen. If a monster dies while you are in the middle of picking items, you will see the new drops. Even if no items are found in the first pass, it tries again just in case.

that's not the entire story.  it generates a new screenshot until no items with any template matches **OR** it hits the time out.  if it is going to timeout, which it will easily and frequently do in the situations i have mentioned, it will simply take what is near (label-wise near) which could be gold and leave valuable loot.  i know this is true because i've seen it and it's clearly that way in the code.

> This is too expensive. Pickit processing already takes a long time, and affecting _every_ use of pickit throughout the entire bot by tripling the calls to item_finder for an edge case would be too much to ask. Part of the difficulty with this is that there are several areas where pickit is called while you are still potentially being hit by monsters and would greatly increase the risk of death/chicken.

the characters (two at the same time) that i bot are poor man's sorceress' that are fragile to say the least and they rarely are killed while looting except on nil and this change hasn't made that any worse.  

> I don't see how prioritizing items in the fashion you have solves the problem you describe. Let's say you have a HR drop, but that HR doesn't have an item label because there are too many stacks of gold/pots on the ground. Now you run pickit. Item_finder finds THE SAME items regardless of what you prioritize.
> 
> * If you prioritize gold first, it'll pickup all the gold and hopefully the HR will come into view.
> * If you prioritize HR first, then pots, then gold; you will skip HR, pick pots, and still pick the same gold. The HR still comes into view.
> * Botty, as it already does, will just check for any pickable item on the screen and pick the closest one, take a new screenshot, pick the next closest, etc. until all the closest drops are picked and will eventually see the HR.
> 
> You end up picking the exact same items in all 3 cases and in all 3 cases you get the HR

it doesn't pick up the same items at all.  as it stands today the bot will frequently hit the looting timeout and just leave high quality loot if it didn't start near (label-near) the high quality loot.  i mention "label-near" again because the label of an item right at the bot's feet may actually be an item at the edge of the screen and in fact a label at the very edge of the screen may be the closest item to the bot.  the bot can and does teleport away to pick up items because its label was nearer even when high quality items are within reach without teleport.  this may actually put good items closer or farther way depending on layout and how d2r deals with label clipping.

anyway, this PR addresses the issue of leaving **seen** loot behind.  regardless of the players misc_gold setting in pickit the first and third calls strip all gold from the item_lists.  the bot always has time to pick the high quality loot before the timeout because it doesn't share a timeout with gold.  this is because i moved gold to its own timeout (the second call).

as i mentioned previously, during the first pass the bot strips gold and is allowed to pick only pots and items.  this allows the bot to pull any important items and potions if it is still getting hit.  then, if the player has gold pickup enabled it does a gold pass on a separate timeout.  it will hit this timeout in the scenarios i have provided nearly always.  instead of quitting back to town it does another (third) pass that strips gold and pots just in case the gold pick made more labels appear.

you aren't considering timeouts in any of your scenarios and it is the timeouts that are killing the capability of picking the quality item in the situations i have outlined.  this PR gives you the ability to tweak timeouts and ensures that items are on a different timeout than the gold and pots at least once.

if the PR can't pass as is then i can implement something definitely cheesier where during the single timeout at certain amounts of passed time i switch between clearing/ignoring gold or potions so that by the end it's never allowed to consider gold.

at any rate, the pr does solve the issue i outlined albeit expensive in your opinion.  i think much more expensive is leaving behind high runes.  also, i think it's hardly fair to call this an edge case.  a non-infinity sorceress (and i think all the generic caster classes without infinity which is the bulk of the characters in the game) can run 6 or fewer scripts currently; nihl, pindle, eldritch, shenk, trav, and arcane.  my blizz sorc can't do shenk and my fake lightning sorc can't do trav so really and neither one reliably succeeds against nihl.  so really they do 4.1 scripts successfully (about a 10% success rate on nil but they also fail on the others often enough).  

arcane is really up to 4 scripts/loots in one because it runs all cardinal directions until it find the pedestal platform.  you can barely say that 1 in 5 is an edge case and you definitely can't say that 4 in 8 or 4 in 9 is an edge case.  if it helps 1 in 5 chances to loot i think that's kind of a big deal.  if it helps 4 in 8 chances to loot i think that's a really big deal.

mgleed:
@codahq 

I think that prioritizing higher value items is a reasonable idea. For one, picking up the most valuable items in view would be useful as if you need to chicken or die at least you'll have picked up the most worthwhile things. 

I don't agree with your solution though. As I see it, you can still prioritize items by sorting the results of item_finder without having to call item_finder.py to segment every screenshot multiple times and run through every item template for each item contour needlessly. Just adjust the logic of these lines in pickit.py:

```
            else:
                found_nothing = 0
        ---> add logic to sort item_list here, make the highest priority items earliest in the list.
                closest_item = item_list[0]
                for item in item_list[1:]:
                    if closest_item.dist > item.dist:
                        closest_item = item
```

The other thing you're asking for is to increase the pickit timeout. This will also have dangerous side effects as the pickit timer is there for a reason. If you get stuck during pickit at all (unreachable items at trav, walking characters, inventory filled but not detected, etc.) this will ruin run efficiency and cause more chickens, deaths. Do you have any creative solutions to address both scenarios?

I won't plan to merge as-is, but if @aeon0 wants to review this PR and give his thoughts that'd be welcome.

codahq:
would you merge this then?

```closest_item = item_list[0]
for item in item_list[1:]:
    if (item.name != "misc_gold" and closest_item.name == "misc_gold") or (closest_item.dist > item.dist and (item.name != "misc_gold" or closest_item.name == "misc_gold")):
        closest_item = item```

always replace gold with non-gold or look at the distance as long as closest item isn't replaced by closer gold.  this gets my goal accomplished as well.

PR #488, opened 2022-01-28
Nihlathak spelling

mgleed:
Nobody knows or cares how Nihlathak is actually spelled, which is fine, but there are three different spellings for it throughout the code. (Nihlatak, Nilathak, and correct Nihlathak). Spelling may become more important once OCR is used on things in UI like resistances, names, etc. Updated spelling and tested.


egut125:
Looks good! Not able to currently test it, but can run some nihl runs tomorrow morning to make sure things still function if needed

egut125 (approved):

mgleed:
> Looks good! Not able to currently test it, but can run some nihl runs tomorrow morning to make sure things still function if needed

Thanks! Shouldn't be necessary--ran a few times without issue including from other acts.

PR #491, opened 2022-01-29
Implement Gem Transmute

VladimirMakaev:
Added basic ability to transmute flawless gems into perfect gems. It can work as stand alone script by running:
```
python src/gem_transmute.py
```
It's also integrated along side stashing routine and will perform transmute once every X game (configurable)

See **_transmute.ini_** for new transmute specific settings. I don't really have preference whether to keep it separate or to append those settings to **_params.ini_**

I've separated the picking algorithm from the UI interactions and was able to cover it with unit tests.

Recording of the transmute session:
[https://streamable.com/dprj9m](url)


config/game.ini
```diff
@@ -121,10 +121,13 @@ no_pickup=5,530,240,65
 repair_needed=1000,0,280,280
 wp_act_roi=42,66,365,53
 stash_btn_roi=20,52,412,53
+hero_stash_roi=35,92,378,378
```
mgleed, line 124: `hero_stash_roi` and `vendor_stash` should be equivalent. I'd pick one rather than have two. Personally I'd consider making generic names like `left_inventory` for left panel and `right_inventory` for right panel inventory or something that can be used in vendor routines and stashing routines and change existing names accordingly. Just my 2 cents :)

mgleed:
Let me know if this is stable on master and I'll plan to merge tomorrow if so

VladimirMakaev:
Rebased on top of the master, retested both gem_transmute.py script and in-game routing

Good to go from my end

mgleed (approved):

PR #492, opened 2022-01-29
Allow color_filter to range <0, >180 in H; etc.

mgleed:
A few misc utils updates:

**color_filter**
Updated logic to allow input of HSV range where H spans over 180 or less than 0 to make red ranges easier. Updated logic in belt_manager accordingly for health potions. Have plans to use this to detect reds slot backgrounds in inventory.

**trim_black**
Utility that will trim black border around an image, similar to photoshop's trim function. Returns the new image and the roi of the cropped area w/ respect to original image.

**mask_by_roi**
Given an image and a region of interest, will either apply a mask to the area surrounding the roi (default, type="regular") or mask out the roi and leave the surrounding image intact (type="inverse"). Returns the mask.


mgleed:
moved the image erosion function from item_cropper.py to misc utils

aeon0 (approved):

Issue #489, opened 2022-01-28
Refactor Epic


aeon0:
Miss use the issue/bug report here for a discussion on refactoring just sharing what I am planning. Its work in progress:

## Removing features
Plan is to remove features that are either not used or "switched" by a config. One example is:
- Searching window offset only with win32 not by templates anymore (find_window_via_win32_api)
- Not the biggest fan of the whole specific discord messaging. I like the generic text-based messaging. But might the minority on this one. (message_api_type)

Looking for input of others that you think we can do without. A few others I would look into are these:
message_highlight=1
d2r_windows_always_on_top=0
graphic_debugger_layer_creator=0

## Split ui_manager.py
Currently ui_manager.py has many functions that are actually functional and dont need any dependency of the instance. Split up ui_manager in different parts: e.g. one might be a stashing.py. This way we are able to remove a ui_manager.py dependency on some.

## Dependency Interfaces
We have sometimes big constructors with repetitive arguments. Espcially on the interface classes. Creating dependency classes will make this a bit more bearable. E.g. for IAct this could look like this:
```python
class IActDependencies(dataclass):
    screen: Screen
    template_finder: TemplateFinder
    pather: Pather
    char: IChar
    npc_manager: NpcManager
```
## General stuff
Find functional things that are currently "trapped" in classes and free them and maybe we can get rid of some dependency injections on some instances this way.

Check again if some implementation details can be separated from control flow. e.g. bot.py should hide all implementation details.

VladimirMakaev:
To the point of ui_manager split the following could be an option

## Composable Screen Objects
### Goal:
Create higher level object interfaces for interacting with various objects in the game. E.g. MainMenu, Stash, Inventory, Cube, Keybinds, Waypoints

### Context:
At the moment various functions are spread around the codebase and mixed with lower-level coordinate conversion code

### Examples:
These are a few ideas how this could look like. But we clearly can define an encapsulation boundary between UI routines and UI elements. Having the ability to decouple one from another
1. Creating a game could look like:
```python
main_menu = SceenObject.waitFor(MainMenu)
# MainMenu would have a “locator” inside pointing to an asset to search for
act = main_menu.select_difficulty(“Hell”)
```
2. Interacting with inventory
```python
act = Act.from_location(location)
chest = act.open_chest()
inventory = ScreenObject.waitFor(Inventory)
while True:
   col, row = inventory.get_loot_area().next_item()
[...]

PR #494, opened 2022-01-29
Add ability to launch botty from either main menu or town

VladimirMakaev:
The benefit of this change is to:
1) improve the development inner loop
2) improve the experience of tweaking settings, e.g. you don't need to "save exit" if you're experimenting with settings

Additionally I'm removing find_window_via_win32_api and making this behavior default. As per conversations in 
#489 and #453


mgleed:
> The benefit of this change is to:
> 
> 1. improve the development inner loop
> 2. improve the experience of tweaking settings, e.g. you don't need to "save exit" if you're experimenting with settings
> 
> Additionally I'm removing find_window_via_win32_api and making this behavior default. As per conversations in #489 and #453

Looks great, thanks!

VladimirMakaev:
Fixed merging conflicts

src/game_controller.py
```diff
@@ -36,15 +36,7 @@ def __init__(self):
         self.bot = None
         self.char_selector = None
 
-    def run_bot(self, pick_corpse: bool = False):
-        if self._config.general['restart_d2r_when_stuck']:
```
mgleed, line 40: After moving this to Bot(), if the d2 restarts, the saved char template does not persist.

mgleed (changes requested):

src/bot.py
```diff
@@ -192,15 +195,50 @@ def is_last_run(self):
...
+                "MAIN_MENU_TOP_LEFT","MAIN_MENU_TOP_LEFT_DARK"
+            ],
+            "start_from_town": [
+                 "A5_TOWN_0", "A5_TOWN_1",
+                 "A4_TOWN_4", "A4_TOWN_5",
+                 "A3_TOWN_0", "A3_TOWN_1",
+                 "A2_TOWN_0", "A2_TOWN_1", 
+                 "A2_TOWN_10","A1_TOWN_1", "A1_TOWN_3"],
+        })
+        match = self._template_finder.search_and_wait(list(transition_to_screens.keys()), best_match=True)
```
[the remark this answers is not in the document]
VladimirMakaev: I actually followed the approach used before. E.g. previously you had this 
```
        self._template_finder.search_and_wait(["MAIN_MENU_TOP_LEFT","MAIN_MENU_TOP_LEFT_DARK"], roi=self._config.ui_roi["main_menu_top_left"])
```
waiting for the main menu to appear indefinatelly. I'm not sure if timing out is a good idea, but maybe I'm wrong. Certainly easy to add, just let me know your thoughts.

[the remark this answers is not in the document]
mgleed: > I actually followed the approach used before. E.g. previously you had this
> 
> ```
>         self._template_finder.search_and_wait(["MAIN_MENU_TOP_LEFT","MAIN_MENU_TOP_LEFT_DARK"], roi=self._config.ui_roi["main_menu_top_left"])
> ```
> 
> waiting for the main menu to appear indefinatelly. I'm not sure if timing out is a good idea, but maybe I'm wrong. Certainly easy to add, just let me know your thoughts.

Yeah, I think it's reasonable to wait indefinitely, but might have to adjust something somewhere else. I wonder how we could improve this. I had a game that didn't properly quit even though it thought it had after max game timer expired. It said "game ended" or whatever even though char was still in game. Ignored max game time and then proceeded to here.

VladimirMakaev:
@mgleed 
The issue was in a TemplateFinder having an old instance of screen and after restart D2R window might get another position so it needs to be refreshed. I wonder how it worked before like that.

src/game_controller.py
mgleed: Do you think this issue can be solved? This breaks the save game recovery if for any reason the character you want isn't the top-most online character. I have already implemented improvements to the character selector, but I don't know how to make the result of CharSelector.save_char_template() persist after a new instance of Bot() is created. Is there a way to save the template in game_controller.py's scope?

VladimirMakaev: I see what you mean now. I'll check later.

VladimirMakaev: I've move it to a class variable. I've modified the __main__ code in char_selector.py. It's able to select the character after I moved the window in between the instances of CharSelector. It's not the greatest way to solve this problem but at least it's encapsulated enough for us to easily change this in the future

mgleed (approved):

mgleed:
I think this is probably good-to-go, tested the next-to-most-recent commit. I'd like aeon to be the one to do final review/merge though since is has some more fundamental codebase changes than most PR's.

src/config.py
```diff
@@ -174,7 +174,7 @@ def load_data():
             "info_screenshots": bool(int(Config._select_val("general", "info_screenshots"))),
             "loot_screenshots": bool(int(Config._select_val("general", "loot_screenshots"))),
             "d2r_path": Config._select_val("general", "d2r_path"),
-            "restart_d2r_when_stuck": bool(int(Config._select_val("general", "restart_d2r_when_stuck"))),
```
aeon0, line 177: Why you dont like trailing `,` :( 
Its nice when a new param is added there will be just one line change instead of two and less risk of merge conflicts of others. It's not critical and probably doesn't happen very often that it really helps, still doesn't hurt to do.

src/bot.py
```diff
@@ -222,6 +261,7 @@ def on_create_game(self):
             else:
                 Logger.error("Failed to detect if /nopickup command was applied or not")
         self.trigger_or_stop("maintenance")
+
```
aeon0, line 264:
```suggestion
```

src/bot.py
```diff
@@ -192,15 +196,50 @@ def is_last_run(self):
...
     def on_create_game(self):
-        keyboard.release(self._config.char["stand_still"])
         # Start a game from hero selection
-        self._game_stats.log_start_game()
         self._template_finder.search_and_wait(["MAIN_MENU_TOP_LEFT","MAIN_MENU_TOP_LEFT_DARK"], roi=self._config.ui_roi["main_menu_top_left"])
         if not self._ui_manager.start_game(): return
-        self._curr_loc = self._town_manager.wait_for_town_spawn()
+        self.trigger_or_stop("start_from_town")
 
+    def on_town_init(self):
```
aeon0, line 240: Rename to `on_start_from_town` please to keep naming convention constant. Or rename the state from start_from_town to town_init.

src/bot.py
```diff
@@ -192,15 +196,50 @@ def is_last_run(self):
...
+                 "A5_TOWN_0", "A5_TOWN_1",
+                 "A4_TOWN_4", "A4_TOWN_5",
+                 "A3_TOWN_0", "A3_TOWN_1",
+                 "A2_TOWN_0", "A2_TOWN_1", 
+                 "A2_TOWN_10","A1_TOWN_1", "A1_TOWN_3"],
+        })
+        match = self._template_finder.search_and_wait(list(transition_to_screens.keys()), best_match=True)
+        self.trigger_or_stop(transition_to_screens[match.name])
+    
+    def on_select_character(self):
```
aeon0, line 223: This will select char now every time a new game is created. Meh : /
We probably want to have some function in char_selector that detects if the correct char is already highlighted...

src/bot.py
```diff
@@ -192,15 +196,50 @@ def is_last_run(self):
...
+                result[asset] = key
+        return result
+
+    def on_init(self):
+        keyboard.release(self._config.char["stand_still"])
+        transition_to_screens = Bot._rebuild_as_asset_to_trigger({
+            "select_character": [
+                "MAIN_MENU_TOP_LEFT","MAIN_MENU_TOP_LEFT_DARK"
+            ],
+            "start_from_town": [
```
aeon0, line 213: We have now 2 places where we have to take care of the list of "town templates". Maybe provide a getter in the town manager to get they asset names and use the getter here.

aeon0 (changes requested):
Sry for the late review. A few more suggestions.

src/ui/char_selector.py
```diff
@@ -6,26 +6,27 @@
 
 
 class CharSelector:
+    _last_char_template = None
```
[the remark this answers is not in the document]
mgleed: This was his solution to a prior discussion we had about this PR breaking character recovery after restart since moving the logic from game controller to bot but then creating a new instance of bot on game restart.

src/config.py
VladimirMakaev: It's not a thing in the language I've been working for over a decade :). But I see your point

src/bot.py
mgleed: I have this already developed. Checks for highlighted box, crops out unique character name. Then can just template match from then on. I can either make a PR after this is merged or add to this PR once everything's good to go.

src/ui/char_selector.py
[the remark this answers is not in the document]
VladimirMakaev: It's a bit nasty to store it like that but we need to store it among Bot instances. I can create a separate singleton Storage object but I think  it's reasonably encapsulated and should fit the purpose for the time being

src/bot.py
VladimirMakaev: I think I didn't change this behavior in my PR and kept it as it was. So I suggest adressing this in a separate PR

VladimirMakaev:
Addressed comments raised by @aeon0

src/bot.py
mgleed: Previously it only selected character after a d2 restart and if a template had been previously saved, so if this PR causes a selection every game creation then that's a change

VladimirMakaev: @mgleed Do you want me to add you as collaborator on my fork for this fix? or you can rebase on this in a separate PR? Or you can give me a patch and I'll apply and test? Let me know what works best for you to get your fix on top of this

mgleed: resolved with commit [a2412c0](https://github.com/aeon0/botty/pull/494/commits/a2412c05a6390e414292339c0fc169209a993de0)

VladimirMakaev:
Fixed conflicts, did a smoke test

PR #499, opened 2022-01-31
Updates to template_search, trim lots of code throughout

mgleed:
**Template_finder**
-Added support for list of templates (not just template names) as input to search()
-Renamed matched template attribute `.position` to `.center` to better describe the value 
-Now returns attribute `.region` that includes the bounding box for the matched template. (plans for this later). Was going to name `.roi` but wanted to disambiguate from rois used *for* template matching.
-Create `templates_gray` array only if `use_grayscale` is True.
-Rescale input image only if `scale != 1`
-Simplified calculation of `ref_point` and `rec`. Moved some of this calculation to `roi_center` in `misc.py` (plans for this later)
-added suppress_debug as an input param to `search_and_wait` suppress the "waiting for templates" log
-added `normalize_monitor` as an input argument to `search_and_wait`.

**Everywhere else**
-Trimmed a lot of `convert_screen_to_monitor()` function calls by just passing `normalize_monitor=True` to template_finder. Cut some lines by not having to reassign this to x_m, y_m.


aeon0 (approved):
lovely

PR #502, opened 2022-02-03
prioritize runes/uniques/set items then pick others

mgleed:
Thank you @codahq #485 

Thought to rename `closest_item` since it will now better represent closest desired item, but there are other variable names in there already like `desired_item` etc so I decided to leave it alone.

### DIFF
--- a/src/item/pickit.py
+++ b/src/item/pickit.py
@@ -79,10 +79,7 @@ class PickIt:
             else:
                 found_nothing = 0
                 item_list.sort(key=itemgetter('dist'))
-                for pattern in ["rune_", "uniq_", "set_"]:
-                    closest_item = next((obj for obj in item_list if pattern in obj["name"]), None)
-                    if closest_item:
-                        break
+                closest_item = next((obj for obj in item_list if "misc_gold" not in obj["name"]), None)
                 if not closest_item:
                     closest_item = item_list[0]
 
--- a/src/item/pickit.py
+++ b/src/item/pickit.py
@@ -1,6 +1,7 @@
 import time
 import keyboard
 import cv2
+from operator import itemgetter
 
 from utils.custom_mouse import mouse
 from config import Config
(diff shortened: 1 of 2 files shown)


codahq:
this still weights gold as high as something like a 4so eth giant thresher or 40@ sacred target and it shouldn't.  a 45@ sacred targe is worth many high runes.  this only sort of addresses the issue.  for those of us with squishy characters we have to run with gold pickup on or the bots run out of gold and can't resurrect mercs or buy potions.  this still leaves valuable loot behind if too much gold exists.  you didn't deprioritize gold.

src/item/pickit.py
```diff
@@ -77,10 +78,13 @@ def pick_up_items(self, char: IChar, is_at_trav: bool = False) -> bool:
...
-                for item in item_list[1:]:
-                    if closest_item.dist > item.dist:
-                        closest_item = item
+                item_list.sort(key=itemgetter('dist'))
+                for pattern in ["rune_", "uniq_", "set_"]:
+                    closest_item = next((obj for obj in item_list if pattern in obj["name"]), None)
+                    if closest_item:
+                        break
+                if not closest_item:
+                    closest_item = item_list[0]
```
egut125, line 87: Doesn't seem like were doing to closest check on things that don't match the patterns anymore? Not sure if that matters or not as probably less important after the pattern items are grabbed.

Otherwise looks good to me

egut125 (approved):

egut125:
> 

Can we do a reverse of this for items with misc_ in the name to grab those last? Or maybe just misc_gold as the last thing I guess?  
I never really noticed a huge issue with chickening while pickit-ing and can't really see a cost effective way to do this unless we want to add something to pickit ini for order preference but also seems excessive.

codahq:
> Can we do a reverse of this for items with misc_ in the name to grab those last? Or maybe just misc_gold as the last thing I guess?

yes, this is all i really wanted.  deprioritize gold.  make it absolutely last priority.  i think we should do the reverse of this and just deprioritize gold.

src/item/pickit.py
mgleed: Yeah since the list is sorted by distance before the loop it'll be fine

codahq:
perfect.  this will absolutely be better.  thank you.

mgleed:
Tested and works well. Ty @codahq @egut125

PR #503, opened 2022-02-03
New Param for fails, if max is reached, kill botty. Defaults off.

egut125:


mgleed (approved):
To the point! Looks good to me

PR #505, opened 2022-02-04
OCR as a service

mgleed:
Added ocr.py and a few tesseract models for people to experiment with if desired while we develop the pickit config, etc. Will be extending other PR's from this.


egut125 (approved):
Seems reasonable. Not terribly in-depth review but general flow seems good and results are good. Good to get in and iterate as needed.

VladimirMakaev:
I'd land this sooner rather than later. Even if it's not perfect since nothing would break as it's just a service. The only I'd do is go through API of the OCR and prefix methods that are supposed to be private with _.

PR #500, opened 2022-02-01
Messaging-Update-2-message-harder #2

mgleed:
Had some difficulty merging #430 so I created a new branch & PR. Credit to @skizotenigma.

PR #510, opened 2022-02-04
Added exception handling to Discord API calls

pokzcodes:
Last week when Discord's servers were down, I discovered that there were uncaught exceptions being thrown in the Discord API logic that caused Botty to crash, I was botting on a Hardcore character at the time, and this left my character standing in-game at the Waypoint without Botty running (no chicken). This could cause a Hardcore death.

This PR added exception handling to the Discord API calls that have the potential to throw exceptions, possibly preventing unnecessary deaths in the event of a Discord outage or Discord Webhook configuration errors.

### DIFF
--- a/src/messages/discord_embeds.py
+++ b/src/messages/discord_embeds.py
@@ -1,19 +1,21 @@
 from .generic_api import GenericApi
 from config import Config
+from logger import Logger
 import cv2
 import datetime
 import traceback
 import discord
 from version import __version__
 import numpy as np
-from discord import Webhook, RequestsWebhookAdapter, Color
+from discord import Webhook, RequestsWebhookAdapter, Color, InvalidArgument
 
 class DiscordEmbeds(GenericApi):
     def __init__(self):
         self._config = Config()
-        self._webhook = Webhook.from_url(self._config.general['custom_message_hook'], adapter=RequestsWebhookAdapter(), )
-        self._file = None
-        self._psnURL = "https://i.psnprofiles.com/games/3bffee/trophies/"
+        try:
+            self._webhook = Webhook.from_url(self._config.general['custom_message_hook'], adapter=RequestsWebhookAdapter(), )
+        except InvalidArgument:
+            Logger.warning(f"Your custom_message_hook URL {self._config.general['custom_message_hook']} is invalid, Discord updates will not be sent")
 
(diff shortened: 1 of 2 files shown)


mgleed:
Thanks for doing this @pokzcodes . We may be merging a messages update #500  prior to this one so you might have to do some merging, just a heads up for when we do.

src/messages/discord_embeds.py
```diff
@@ -1,19 +1,21 @@
...
 import numpy as np
-from discord import Webhook, RequestsWebhookAdapter, Color
+from discord import Webhook, RequestsWebhookAdapter, Color, InvalidArgument
 
 class DiscordEmbeds(GenericApi):
     def __init__(self):
         self._config = Config()
-        self._webhook = Webhook.from_url(self._config.general['custom_message_hook'], adapter=RequestsWebhookAdapter(), )
-        self._file = None
-        self._psnURL = "https://i.psnprofiles.com/games/3bffee/trophies/"
```
egut125, line 16: Why removal of these two variables?

src/messages/generic_api.py
```diff
@@ -42,4 +44,7 @@ def _send(self, msg: str):
 
         data = json.loads(self._config.advanced_options['message_body_template'].format(msg=msg), strict=False)
 
-        requests.post(url, headers=headers, json=data)
+        try:
+            requests.post(url, headers=headers, json=data)
+        except BaseException as err:
+            Logger.error("Error sending generic Discord message: " + err)
```
egut125, line 50: Not necessarily a discord message, update logger message to some like generic message? Little nit-picky I know

egut125 (changes requested):

src/messages/discord_embeds.py
pokzcodes: @egut125 Ah not sure how those lines got deleted. I just pushed a commit to restore them. Good catch.

src/messages/generic_api.py
pokzcodes: @egut125 What do you mean exactly?

egut125: Remove the word discord from the string. Since it could be used for something other than discord.

pokzcodes: @egut125 Ah, gotcha. Done.

egut125 (approved):

PR #511, opened 2022-02-04
Added logic to kill D2R when Botty gets stuck and shuts down (prevents issue that causes HC deaths)

pokzcodes:
I've lost two HC characters to Botty getting stuck and shutting down, leaving them in-game to die with no chicken protection. 

In this PR, I added

- a method in restart.py to kill D2R
- a method to safely quit Botty by first killing D2R before calling os.exit
- a call to the safe_quit method during a specific instance in the code that has killed two of my HC bot characters


mgleed (approved):
Agree with the need for this. Code looks good and you confirmed you tested. Thank you

PR #513, opened 2022-02-05
Bugfix for TP charge detection: use color_filter instead of template match threshold

mgleed:
Fixes bug reported by a couple discord users who had <95% match for assets


src/char/i_char.py
```diff
@@ -34,11 +35,13 @@ def __init__(self, skill_hotkeys: Dict, screen: Screen, template_finder: Templat
     def _discover_capabilities(self) -> CharacterCapabilities:
```
VladimirMakaev, line 35: It looks good. But I’m not an expert on how color  filters work. However the main challenge was to distinguish between 22 charges and 2 charges for instance that’s why >95 was used for original low_on_charges tracking. I think there is potentially an issue too

VladimirMakaev (approved):

src/char/i_char.py
```diff
@@ -34,11 +35,13 @@ def __init__(self, skill_hotkeys: Dict, screen: Screen, template_finder: Templat
...
-                    return CharacterCapabilities(can_teleport_natively=True, can_teleport_with_charges=False)
-                else:
+                img = self._screen.grab()
+                skill_img = cut_roi(img, self._config.ui_roi["skill_right"])
+                charge_mask, _ = color_filter(skill_img, self._config.colors["blue"])
+                if np.sum(charge_mask) > 0:
                     return CharacterCapabilities(can_teleport_natively=False, can_teleport_with_charges=True)
+                else:
+                    return CharacterCapabilities(can_teleport_natively=True, can_teleport_with_charges=False)
             return CharacterCapabilities(can_teleport_natively=False, can_teleport_with_charges=True)
```
egut125, line 42: What is this line doing? Seems like extra return?

egut125: Maybe it's mobile formatting. If it's all good this looks good

egut125 (approved):

PR #515, opened 2022-02-05
Bugfix: runs_per_repair modulo by zero fix

mgleed:


egut125 (approved):
Looks good. I've noticed we accomplished this in at least two different ways for different parameters. Might make a unifying pr this week for them.

PR #516, opened 2022-02-05
Track item charges with OCR

mgleed:
Continuation of previous PR in which the native TP vs. charged TP bug was partially fixed. Leverage OCR to read charges remaining on charged skill. Depends on #505 .

Relevant changes are in i_char.py and bot.py

Would appreciate review from @VladimirMakaev


src/ocr.py
```diff
@@ -0,0 +1,290 @@
...
+        check_known_errors: bool = True,
+        check_wordlist: bool = True,
+        word_match_threshold: float = 0.9
+    ) -> list[str]:
+        """
+        Uses Tesseract to read image(s)
+        :param images (required): image or list of images to read in OpenCV format.
+            Use a list of images rather than looping over single images where possible for best performance.
+        :param model: OCR language model basename to use (in assets/tessdata folder)
+        :param psm: Tesseract PSM to use. 7=single uniform text line, 6=single block of text, 3=auto without orientation.
```
VladimirMakaev, line 89: Ideally I wouldn't expose magic numbers outside and would rather create an Enum that describes semantics of each of those PSMs. e.g.
```python
class Segmentation(Enum):
     Auto = 0,
     AutoNoOcrOrOSD = 1,
     ...etc (not sure I correctly made examples but the point is for API to be self-descriptive)
```

src/ocr.py
```diff
@@ -0,0 +1,290 @@
...
+    original_text: str = None
+    word_confidences: list = None
+    mean_confidence: float = None
+    # these are kept to help train OCR
+    original_img: np.ndarray = None
+    processed_img: np.ndarray = None
+    def __getitem__(self, key):
+        return super().__getattribute__(key)
+
+class Ocr:
```
VladimirMakaev, line 26: I'd try to add some encapsulation to this class by prefixing methods which are not intended to be called by others with "_"

VladimirMakaev (commented):
I've tested walkadin with and without charges. All works well. Thanks @mgleed for taking this flaky PR to the next level!

src/ocr.py
mgleed: There already is this in Tesserocr for the OEM and PSM classes but either way you need a reference--either to look up what number to use with the magic numbers or to look up what keywords to put after PSM or OEM. I found it less cumbersome this way personally. As you say though it would make the code more readable.

egut125 (approved):
Looks good!

PR #517, opened 2022-02-06
Fix game isn't restarted when in main menu / character selection screen

VladimirMakaev:
1) game isn't being tracked up until start_from_town which means max_game_length_s isn't accounted for during character selection
2) restart_game utility isn't handling new screen behavior properly and depending on it to wait for main menu while it's not


mgleed (approved):
Thank you

PR #520, opened 2022-02-06
Bugfix: Provide precompiled tesserocr

mgleed:
Tesserocr with Tesseract > 5 works with pip install but installer fails if user doesn't have VS compilers. Will revisit


egut125 (approved):

mgleed:
@egut125 ended up solving this by providing precompiled Tesserocr in the new `dependencies` directory. Tested on another machine without vs tools, worked well. Properly installs with tesseract >= 5.0.1.

PR #521, opened 2022-02-06
Cleanup: Artifact from recent PR

mgleed:


egut125 (approved):
Guessing this file is just no longer needed?

mgleed:
> Guessing this file is just no longer needed?

Yeah I just noticed it was blank and not referenced anywhere.

PR #525, opened 2022-02-07
Parallelize template searching

Ezro:
Proposal to parallelize search_and_wait template waiting using futures.

### DIFF
--- a/src/template_finder.py
+++ b/src/template_finder.py
@@ -93,7 +93,6 @@ class TemplateFinder:
         recs = [[0, 0, 0, 0]] * len(templates)
         future_list = []
         for count, template in enumerate(templates):
-            template_match = TemplateMatch()
             scale = scales[count]
             if scale != 1:
                 img: np.ndarray = cv2.resize(
@@ -128,6 +127,7 @@ class TemplateFinder:
                     break
             # else:
             #     Logger.error(f'{ref[i]} not found')
+        template_match = TemplateMatch()
         if len(scores) > 0 and max(scores) > 0:
             idx = scores.index(max(scores))
             try:
@@ -138,10 +138,6 @@ class TemplateFinder:
             template_match.score = scores[idx]
             template_match.region = recs[idx]
             template_match.valid = True
-        else:
-            template_match = TemplateMatch()
-            # Logger.debug(
-            #     f'Defaulting TemplateMatch result for ref {ref}')
         return template_match
 
     def parse_ref(self, ref, use_grayscale, best_match):
--- a/src/template_finder.py
+++ b/src/template_finder.py
@@ -10,8 +10,10 @@ import time
 import os
 from config import Config
 from utils.misc import load_template, list_files_in_folder, alpha_to_mask, roi_center
+import concurrent.futures
 
 template_finder_lock = threading.Lock()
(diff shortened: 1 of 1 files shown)


VladimirMakaev:
Given the fact that this is CPU bound workload do you have any metrics behind this change? My concern is that we'll have spikes in CPU consumption by botty while it's already mostly run in a constrained environment with many VMs fighting for resources. I may be wrong so some metrics would be nice

Ezro:
> 

I don't have any concrete benchmarks on hand, but when I developed this I noticed that the initial A5_... searching took >1s for multiple searches (until found). With the current implementation in main, this also will run into scalability issues if you ever wanted to match against larger sets.

From my testing, those calls were reduced from >1s to ~0.2s (if memory serves).

The PR includes debug lines for printing the elapsed time for image grabbing as well as searching; I'd encourage pulling down the branch and messing around with the before vs after.

Ezro:
> I completely understand that you'd decrease the search latency as you'll utilize more CPU cores. My concern is about how it would affect setups with VMs and whether you won't overcommit your CPU or create more issues in constrained environments. Ideally the amount of threads you create should be capped somehow based on the amount of cores. I'm not sure if ThreadPoolExecutor would do it natively, let me know if it does
> 
> Also I don't think Futures is the right abstraction for the problem and I'd rather go for a solution with "Parallel-For" semantics that gives an ability to limit concurrency and has a much cleaner syntax with exactly the same outcome. That would allow to limit your concurrency to 1 by a setting a parameter in config as well
> 
> But I'd ask @aeon0 to review this anyway as those are just my thoughts

That's fair. I don't use a VM for my setup so I'm unable to test the solution against it.

From the [Python docs](https://docs.python.org/dev/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor):
`If max_workers is None or not given, it will default to the number of processors on the machine, multiplied by 5, assuming that [ThreadPoolExecutor](https://docs.python.org/dev/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) is often used to overlap I/O instead of CPU work and the number of workers should be higher than the number of workers for [ProcessPoolExecutor](https://docs.python.org/dev/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor).`

We could manually set the max_workers, but I'm not convinced that the default behavior would be detrimental to a VM.

re: Futures vs Parallel-For (multiprocessing?), I think this is more of a subjective thing; imo, futures should be the go-to for the majority of parallelized tasks since it has a simpler API and my understanding is that it was designed specifically as an easier-to-use replacement for existing concurrent libs (like multiprocessing).

Ezro:
Main CPU Benchmark
Parallel Search CPU Benchmark

VladimirMakaev:
It doesn't look bad based on the benchmark data. Thanks for getting this. Main folks who probably wouldn't like this are the ones running 3 VMs but if the bottleneck is GPU rather than CPU it can work out just fine.

mgleed:
@Ezro if you're still interested in this PR, see if you can rebase/merge given recent template_finder changes. Will review again at that point. Thanks!

jobithu:
Imagine what we could do with our mob detection if this thing here was working :)

mgleed:
Closing for now, can revisit in the future if need arises

PR #528, opened 2022-02-09
Update transmute.py

mgleed:


VladimirMakaev (approved):
lgtm

egut125 (approved):
No longer worth a threshold check?

mgleed:
> No longer worth a threshold check?

The ROI is small so there shouldn't be any false positives. One user was having issues and fixed them by recapturing his own transmute asset--guessing this has to do with yet-to-be-explained template matching variation.

PR #531, opened 2022-02-09
Add Horker Barbarian Weapon Switch by a user

jobithu:
- PR'd by me on behalf of a user
- adds a simple weapon switch before horking.
- after horking, weapon is switched back to main hand.
- no check for success implemented - could be a next step (copy code from CTA pre-buff routine to safeguard against issues whilst swapping weapons).

### DIFF
--- a/src/char/barbarian.py
+++ b/src/char/barbarian.py
@@ -8,10 +8,13 @@ from logger import Logger
 from screen import Screen
 from utils.misc import wait, cut_roi
 import time
+import os
 from pather import Pather, Location
 
 
 class Barbarian(IChar):
+    os.system('color')
+    
     def __init__(self, skill_hotkeys: dict, screen: Screen, template_finder: TemplateFinder, ui_manager: UiManager, pather: Pather):
         Logger.info("Setting up Barbarian")
         super().__init__(skill_hotkeys, screen, template_finder, ui_manager)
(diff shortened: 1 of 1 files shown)


src/char/barbarian.py
```diff
@@ -153,7 +189,16 @@ def kill_nihlathak(self, end_nodes: list[int]) -> bool:
...
+        wait(0.5, 0.55)
+        #switch to off hand before horking
+        Logger.info('\033[92m'+"SWITCHING TO OFF HAND!!!!!"+'\033[0m')
+        keyboard.send(self._char_config["weapon_switch"])
+        wait(0.1, 0.15)
+        self._do_hork(4)
+        wait(0.5, 0.55)
+        #switch back to main hand
+        Logger.info('\033[92m'+"SWITCHING BACK TO MAIN HAND!!!!!"+'\033[0m')
+        keyboard.send(self._char_config["weapon_switch"])
```
VladimirMakaev, line 201: should encapsulate weapon switching in the char. E.g. have char.switch_weapon() method that deals with resiliency  of the process. CTA routine should use that one too.

src/char/barbarian.py
```diff
@@ -137,7 +164,16 @@ def kill_council(self) -> bool:
             # Stay inside and cast war cry again moving forward
             self._move_and_attack((40, 10), atk_len)
             self._move_and_attack((-40, -20), atk_len)
-        self._do_hork(5)
+        wait(0.5, 0.55)
```
VladimirMakaev, line 167: can we further encapsulate this in a form of:
```python
self._do_hork(5, offhand=True)
```
Can hide shenanigans with waiting and swapping and swapping back there

VladimirMakaev (changes requested):

jobithu:
closing, as tw1st has a better solution available already

PR #534, opened 2022-02-11
Track nopickup status in GameStats and only /nopickup once

Ezro:
While the initial code was more resilient, this approach aims to:
* Reduce visual noise (if you're watching the bot)
* Reduce possible behavioral pattern detection


mgleed (approved):
Looks good, thank you

PR #541, opened 2022-02-12
Break Time

Cathb0nes:
**take_break** = amount of games before breaking, can be left blank to never break
**take_break_time** = value in seconds to break for


src/config.py
```diff
@@ -221,6 +221,8 @@ def load_data():
             "stash_gold": bool(int(Config._select_val("char", "stash_gold"))),
             "gold_trav_only": bool(int(Config._select_val("char", "gold_trav_only"))),
             "use_merc": bool(int(Config._select_val("char", "use_merc"))),
+            "take_break":  False if not Config._select_val("char", "take_break") else int(Config._select_val("char", "runs_per_repair")),
```
mgleed, line 224: review this line

PR #542, opened 2022-02-12
Breaktime

Cathb0nes:
corrected bad values in config.py


Ezro:
I wonder if the implementation should have a random element to it. For instance, the config could have something like an ideal_runtime_hours and ideal_break_minutes. The reason, I'm thinking, is for anti-behavioral pattern detection.

With those set, the bot may run for the specified time the first iteration, but then the second iteration could vary within reasonable parameters.

e.g.,
ideal_runtime_hours=4
ideal_break_minutes=15

```
Iteration 1:
* runtime: 4 hours
* break time: 15m
```

```
Iteration 2:
* runtime: 3.5 hours
* break time: 30m
```

```
Iteration n:
* runtime: 1.5 hours
* break time: 10m
```

Another aspect, in a similar vein, is detecting if it's "overrunning" (i.e., no sane human would run for 4 hours, sleep 30m, run 4 hours, sleep 30m, run 4 hours, sleep 30m, etc.)

This isn't to say that these two implementations are mutually exclusive.

PR #545, opened 2022-02-14
Fix incorrect path offset for shenk. Optimize tele charges

VladimirMakaev:
The offset is actually required in both cases (tele or walking) as the point is to get closer to shenk

I've also forced using tp as normally pre_move is responsible for that but we want to keep the concentraction up if we move on teleporting character

Additionally enabled tp_charges on shenk and eldrich routes to take advantage

Tested:

- Hammerdin with enigma
- Hammerdin walking
- Hammerdin walking with charges


mgleed (approved):
Thank you, genius save!

PR #546, opened 2022-02-14
Sell 'junk' during maintenance

Ezro:


mgleed (commented):
I think this functionality is good, but I think for real-world purposes we should try to consolidate vendor trips. For example, it currently will attempt to heal/buy pots at vendor -> ID -> stash -> sell items at a vendor -> repair and buy tp's. Perhaps we could adjust the logic so if there's a need to go to a vendor for any reason then it should sell junk while at those vendors as to not make an extra trip. 

Keep in mind though that the ultimate plan is to revamp towning/stashing. First, inspect inventory if items are picked, ID if id_items=True and if tome in inventory, ID at cain otherwise, then it will return a list of items to sell, keep/stash, or drop. Then it will visit vendors / stash appropriately and buy any needed pots/scrolls/keys from whatever vendors are interacted with. Will also be able to do the same inspect from out of town to drop unwanted items during long runs.

I have a basic implementation of this on ocr branch but it'll require quite a bit of refactoring to separate a PR for it. I think this might do well in the meantime, but if you have any creative ideas for reducing vendor trips with this lmk.

Ezro:
> I think this functionality is good, but I think for real-world purposes we should try to consolidate vendor trips. For example, it currently will attempt to heal/buy pots at vendor -> ID -> stash -> sell items at a vendor -> repair and buy tp's. Perhaps we could adjust the logic so if there's a need to go to a vendor for any reason then it should sell junk while at those vendors as to not make an extra trip.
> 
> Keep in mind though that the ultimate plan is to revamp towning/stashing. First, inspect inventory if items are picked, ID if id_items=True and if tome in inventory, ID at cain otherwise, then it will return a list of items to sell, keep/stash, or drop. Then it will visit vendors / stash appropriately and buy any needed pots/scrolls/keys/repair from whatever vendors are interacted with. Will also be able to do the same inspect from out of town to drop unwanted items during long runs.
> 
> I have a basic implementation of this on ocr branch but it'll require quite a bit of refactoring to separate a PR for it. I think this might do well in the meantime, but if you have any creative ideas for reducing vendor trips with this lmk.

The current implementation will sell junk and then stash (making the stashing theoretically shorter), but I agree that the overall town visit could / should be improved / consolidated.

I see where you're coming from re: revamping towning/stashing and I do have some ideas on it, but I'm planning on creating a separate branch for that.

Ezro:
Closing this as it has been implemented in the ui-overhaul branch

PR #548, opened 2022-02-15
Bugfix: Shenk offset stacking

mgleed:
Had a bug where hammerdin would progressively move further southeast of Shenk during sequential runs. https://www.youtube.com/watch?v=FXxjy9NRLYw

Also reduced shrine checking spam, reworded debug message to clarify purpose of check

Bumped to v0.6.6

### DIFF
--- a/src/char/hammerdin.py
+++ b/src/char/hammerdin.py
@@ -22,6 +22,7 @@ class Hammerdin(IChar):
 
         self._pickit = pickit #for Diablo
         self._picked_up_items = False #for Diablo
+        self._offset_completed = False
 
     def _cast_hammers(self, time_in_s: float, aura: str = "concentration"):
         if aura in self._skill_hotkeys and self._skill_hotkeys[aura]:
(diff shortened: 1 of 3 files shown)


src/char/hammerdin.py
```diff
@@ -55,7 +56,9 @@ def on_capabilities_discovered(self, capabilities: CharacterCapabilities):
         if capabilities.can_teleport_natively:
             self._do_pre_move = False
         #hammerdin needs to be closer to shenk to reach it with hammers
-        self._pather.offset_node(149, (70, 10))
```
VladimirMakaev, line 58: Can we just put in constructor instead?

mgleed: that's where we had it to begin with right? i thought you said it was in the wrong place that's why you moved it

jobithu:
you could replace the offset 149 node with this new node - only for walking chars:

pather.py
152: {'SHENK_11': (261+70, 395+10), 'SHENK_10': (495+70, 421+10), 'SHENK_13': (393+70, -9+10)},

src/char/hammerdin.py
VladimirMakaev: It turned out it doesn't need to depend on capabilities. But also depending on capabilities is wrong cause capabilities can change  (theoretically) and you only set pather once. So putting it back to constructor is better

VladimirMakaev (approved):

PR #551, opened 2022-02-15
Fix Diablo is passed as char to pickit

VladimirMakaev:
Should probably fix this 
image
@jobithu please take a look


mgleed:
> Tested & works on my side - however, my code uses the exact same syntax as used for every other boss. for me it is not clear, why this error occurs.

It's because you pass `self` which is the class you've created, Diablo, rather than the character, hammerdin. What vladmir's done here is essentially what's done in other runs like arcane.py

PR #555, opened 2022-02-16
Speed up save_and_exit

Ezro:


D2RLegit:
Ran this script for a couple hours. It failed to chicken in time (I have low chicken settings) and ended up getting a death before he could chicken and game ended up getting stuck on death screen. Watched this happen 3 times. You might want a recourse if that happens, like a left click and try again or integrate both options and make your attempt first try and the click attempt second try.

Ezro:
> Ran this script for a couple hours. It failed to chicken in time (I have low chicken settings so not scripts fault, although i haven't seen it work yet) and ended up getting a death before he could chicken and game ended up getting stuck on death screen. Watched this happen 3 times. You might want a recourse if that happens, like a left click and try again or integrate both options and make your attempt the first try and the click attempt on the second try.

Hm... that's interesting. On mine (both VM and local), I ran it for about an hour each and it chickened perfectly. I wonder if it's to do with the wait time for different PCs.

At what point was it failing to chicken? It may be worth it to tweak the wait times if it's pressing them too quickly (e.g., accidentally opening the options, or pressing hotkeys before the window is actually open)

Also, re: death screen, I'm not sure how / why it'd get stuck since I think the same thing could happen with main (?)

mgleed:
This is a requested feature so thanks for making it @Ezro . If anyone else can test it out and lmk, that'd be great. I might be able to test it out some in a couple days. 

I'm not sure what caused @D2RLegit to have issues but I agree making a continency function would be good (like, if after keystroke quit attempt the save & exit template is still visible then attempt to click the save and exit template).

Ezro:
@D2RLegit Can you give the latest implementation a go? It uses hardcoded x,y coords and mouse clicks.

D2RLegit:
Yes, the new update fixes the bug. If it fails to chicken in time, botty will still continue as it should. Exiting seems pretty fast and more reliable with fixed coordinates over the existing read image select. Seems good to go.

mgleed (approved):
Thanks!

D2RLegit:
hmmmm i believe this is causing a new error but cant confirm unless i remove scrip to default and could be an issue with master. After i die and chicken fails, it exist game properly but then it will press esc in lobby and show the "did you want to exit diablo 2" button, which then causes botty to timeout.

mgleed:
@Ezro Both @D2RLegit and I experienced a bug with chickening where when you get back to D2 character select it automatically opens the "Do you want to exit d2" prompt and hangs there. A bit puzzling--I don't see a clear reason why this would be the case. Reverted the PR for now til we can pin down why this happened.

mgleed:
Haven't found the best way to address it, but I tracked the extra keystroke to game_controller.py

```py
            # we must be ingame, but maybe we are at vendor or on stash, press esc and look for save and exit btn
            template_match = self._template_finder.search(["SAVE_AND_EXIT_NO_HIGHLIGHT", "SAVE_AND_EXIT_HIGHLIGHT"], self._screen.grab(), roi=self._config.ui_roi["save_and_exit"], threshold=0.85)
            if template_match.valid:
                self._ui_manager.save_and_exit()
            else:
                keyboard.send("esc")
```

PR #557, opened 2022-02-16
Dia & CS Nova/Chain Lightning Sorc

jobithu:
Added pathing & attack sequences for Chain Lightining Nova Sorc.

You need tanky equipment (Vipermagi, Maras, Griffons, 2 Soj, Magefist, Infernostride, Hoto, Phoenix shield (or Stormshield & Pheonix + CTA on offhand-switch) & a Maxblock build (don't use Energy Shield, though - too many mana burners down there).

For sure some room for improvement, but thats more a fun side-project as proof of concept, feel free to add & adapt. In the end: Hammerdin is the way to go for CS.


mgleed:
@jobithu Hey, it's been a while but now master branch is a bit more stable and no planned overhauls. Reviewing PR's. Do you think this one is worth adapting and merging? I can help if you think it is. Lmk

jobithu:
Think we should first add the mob detection abs then add the functionality to this PR before merging

doschhar:
is this build available to run already?

PR #558, opened 2022-02-16
Refactor: UI overhaul

mgleed:
**Summary of changes:**
-Rearranged many UI-related functions across the codebase (particularly in ui_manager.py) into individual components. See `src/ui_components/`. Functions related to components can be called like `if waypoint.use_wp("Arcane Sanctuary"):` or `loading.wait_for_loading_screen(2.0)`
-Removed Screen and TemplateFinder dependencies across the codebase
-Screen functions separated from class
-TemplateFinder made into a singleton
-Added generic ScreenObject UI functions to simply detect UI elements on screen; e.g., `detect_screen_object(ScreenObjects.TownPortal)`, `wait_for_screen_object(ScreenObjects.Loading, 2)`. Also included related functions like `select_screen_object_match(template_match)` to automatically select the match.
-Consolidated assets. Organized templates folder into components. Adjusted some asset ROI's and alpha channels.
-Changed some template searches to grayscale for speed
-Reworked save/exit routine, should now be more consistent and faster with normal exit, chicken exit, and death exit.
-Got rid of unused and consolidated duplicate game.ini parameters
-spelling fix `_last_merc_healh`
-NPC manager class removed and replaced with functions
-D2 window detection now runs on its own thread for semi-live window position offsets.
-Character selection reworked. OCR your selected character for fun (not that accurate, lol)
-Simplified game start and difficulty select
-Added a get experience function, but this is not called anywhere yet
-Reworked corpse pickup. pickup_corpse is no longer a necessary argument/parameter throughout the codebase (bot.py, game_controller.py, runs, etc.)


VladimirMakaev:
@mgleed You're a coding machine dude :)

src/char/barbarian.py
```diff
@@ -46,7 +47,7 @@ def _do_hork(self, hork_time: int):
         if self._skill_hotkeys["find_item"]:
             keyboard.send(self._skill_hotkeys["find_item"])
             wait(0.5, 0.15)
-        pos_m = self._screen.convert_abs_to_monitor((0, -20))
+        pos_m = Screen().convert_abs_to_monitor((0, -20))
```
VladimirMakaev, line 50: Maybe make those functions class level too? Since you initializing Screen it at class level now?

src/bot.py
```diff
@@ -222,18 +226,19 @@ def on_init(self):
     def on_select_character(self):
         if self._config.general['restart_d2r_when_stuck']:
             # Make sure the correct char is selected
-            if self._char_selector.has_char_template_saved():
-                self._char_selector.select_char()
+            if SelectedCharacter.has_char_template_saved():
+                SelectCharacter(self._template_finder).select_char()
             else:
```
VladimirMakaev, line 231:
```suggestion
            selector = SelectCharacter(self._template_finder)
            if selector.has_char_template_saved():
                selector.select_char()
            else:
```

VladimirMakaev: I understand that has_char_templated_saved is class level but in this context it's nicer to use it as instance based but still have your storage at class level

src/char/barbarian.py
```diff
@@ -1,4 +1,5 @@
 import keyboard
+from ui_components.skills import is_left_skill_selected, is_right_skill_active
```
VladimirMakaev, line 2:
```suggestion
import ui_compoents.skills as skills
```

VladimirMakaev: This will allow a usage pattern like
```python
skills.is_left_skill_selected()
```
Much better discoverability of other things you put there over time

src/bot.py
mgleed: got it, changed, thanks

src/char/barbarian.py
VladimirMakaev: Can we apply this to all other usages too?

VladimirMakaev (commented):
I'm not a big fan of Screen().xxx and TemplateFinder(..).yyy everytime. I understand the singleton pattern but isn't it cleaner to just do Screen.xxx and TemplateFinder.yyy instead? I can't think about any immediate benefit of having an instance for both of those classes. The internal state of Config is stored in class level variables, TemplateFinder is just a utility in it's nature as well. There is no harm in leaving this as is if its a big ask, but just raising to discuss

screen_objects.json
```diff
@@ -0,0 +1,80 @@
+{
```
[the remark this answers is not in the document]
Ezro: @VladimirMakaev I disagree with this strongly. I don't think there's any reason to keep Locator or ScreenObject.

With this new approach, if we wanted to add a new screen object then we can add it to our json config and then it will "just work" (i.e., you'll be able to say ui_manager.SCREEN_OBJECTS['MyNewScreenObject'].

With what you guys have, you'd have to create new classes and push code; I don't understand the value add to the existing architecture, but I'm open to discussion.

Ezro:
> I'm not a big fan of Screen().xxx and TemplateFinder(..).yyy everytime. I understand the singleton pattern but isn't it cleaner to just do Screen.xxx and TemplateFinder.yyy instead? I can't think about any immediate benefit of having an instance for both of those classes. The internal state of Config is stored in class level variables, TemplateFinder is just a utility in it's nature as well. There is no harm in leaving this as is if its a big ask, but just raising to discuss

I'm not opposed to that, but I don't really view it as being any different (functionally); we can achieve what you're saying by doing something like:
```
screen = Screen()
screen.grab()
```

Whether we store the singleton in a local variable or use it directly doesn't make a difference imo, but I can see how the code may look cleaner that way.

Another approach would be to break the singleton(s) completely and have everything managed in their respective files.

e.g.,
```
# screen.py
sct = mss()
monitor_roi = sct.monitors[0]

def grab() -> np.ndarray:
    img = np.array(sct.grab(monitor_roi))
    return img[:, :, :3]
```

I'd be in favor of this approach too because then all of the callers would have an even cleaner contract:

```
from screen import grab

def my_func_somewhere_else():
    img = grab()
```

What do you think @VladimirMakaev?


Personally, I like having the Screen() and TemplateFinder() because it (hopefully) reads / signifies that it's a singleton; whereas, if it's stored in a local variable now you have one layer of (possibly unnecessary) indirection.

screen_objects.json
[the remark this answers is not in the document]
VladimirMakaev: This is exactly the opposite from the idea. If that's becomes a dictionary of locators then it's just better to hard code it as such. There is absolutely zero value in having json. 
e.g
```python
class Locators:
     PlayBtn = Locator(...)
     MainMenu = Locator(...)
```
At least you'd get IDE helping you against syntax erros

Original idea was to have objects that encapsulate parts of the ui and have behavior & state attached to those parts of the UI. If we see there is no behavior then just have those in Python. I'm not really sure what you're achieving with your statement of "just work" since you still have to write functions that make sense of those locators and declaring a locator in code or in json it's essentially the same with a downside of not being a part of the language

src/ui/ui_manager.py
```diff
@@ -1,761 +1,150 @@
...
+        ref = screen_object['ref'],
+        time_out = time_out,
+        threshold = threshold,
+        roi = roi,
+        best_match = best_match,
+        use_grayscale = use_grayscale,
+        normalize_monitor = normalize_monitor)
+    if match.valid:
+        return match
+    return match
```
mgleed, line 132: Already mentioned to Ezro, but lets cleanup the redundant code here before merging to master. Whether that means making a @dataclass for the default values, getting rid of template_finder.search_and_wait and just making that a part of ui_manager.py, or whatever it is.

Ezro: This should be resolved with the introduction of ScreenObject / ScreenObjects.

src/char/barbarian.py
mgleed: Fixed

src/utils/misc.py
```diff
@@ -41,18 +41,22 @@ def match(self, hwnd) -> bool:
             result = False
         return result
 
+specs = [
+    WindowSpec(title_regex="D2R1 on .+"),
+    WindowSpec(process_name_regex="D2R.exe")
+]
```
mgleed, line 48: @Ezro Thanks for rebasing. I see you've adapted the changes from #563 but I don't see the new params used anywhere, which I think was the purpose of that PR to allow customization. Would you see if you can incorporate these? The new params were appropriately merged under [advanced_options] in params.ini but don't look like they got put into Config in the rebase and aren't used here. Thanks

src/screen.py
```diff
@@ -4,86 +4,133 @@
...
+def grab() -> np.ndarray:
+    global monitor_roi
+    img = np.array(sct.grab(monitor_roi))
+    return img[:, :, :3]
+
+position = find_d2r_window()
+if position is not None:
+    set_window_position(*position)
+
+# class Screen:
```
aeon0, line 61: Can we remove this commented code?

aeon0 (commented):
I am cool with a bit more static and functional code. And the UI split is nice. I haven't look through every single line of code. When you feel the code is "clean" enough to be merged, feel free to do so.

src/screen.py
mgleed: Yeah there's other places with comments too I'm planning to trim once complete. Thanks!

src/utils/misc.py
Ezro: This should be resolved now. I tried tweaking the implementation during the rebase, but I've reverted it back to how it is in main.

mgleed:
Tested on Ezro's and my machines. Tested chicken, game recovery, and death as well. Working nicely.

Have not yet tested on VM.

Ezro:
> Tested for hours on Ezro's and my machines. Tested chicken, game recovery, and death as well. Working nicely.
> 
> Have not yet tested on VM.

I've tested for 8+ hours on both VM and local.

VladimirMakaev (approved):
Thanks @Ezro @mgleed. Really massive effort to allow everyone to move faster!

egut125 (approved):
Looks good I like the changes nice work everyone!

Issue #562, opened 2022-02-17
Botty No Longer Detects D2R Window When Running on Host and D2R Running in VM


codahq:
**Botty Version**
Current master as of 2022-02-17

**Describe the bug**
Botty can't detect the D2:R window. I reported this would be a problem.  For reference here:
https://github.com/aeon0/botty/pull/453

It is related to that PR and other changes with python using the Win32 api instead of templates.

**To Reproduce**
1. Start botty.  
2. It fails to detect the D2:R window.

**Expected behavior**
It detects the D2:R window.

**Additional info**
@AlekseyLesnoy, @VladimirMakaev, @mgleed  I thought the agreed approach was that the Win32 api would be used first and if it failed it would default back to the old method.  It is not defaulting back to the old method.  Instead, it doesn't find the window at all and cannot start.  I am only finding this now because my VM and other host machine are running compiled versions which I didn't update until today.

Potentially related commits or pulls
https://github.com/aeon0/botty/commit/95c4431206548475c8ec66d9443dd2b6dd46807e
https://github.com/aeon0/botty/pull/483

VladimirMakaev:
@aeon0 decided to remove asset matching support in this commit to simplify the code and also removed monitor configuration requirements

https://github.com/aeon0/botty/commit/8907a46073d65d108f556957247ba227431733e1

codahq:
if the goal of the pixel bot was to be undetectable a vm won't ever be able to detect a process, sdk, or anything on its host.  that's my use case; to be extremely undetectable.  the only option for detection would be forensics on mouse movement and character behavior.  

mouse and keyboard won't passthrough unless you run python/botty from an elevated command prompt so open a command prompt/powershell as adminstrator.

Ezro:
@codahq Not trying to be that guy, but what's the advantage / point to sending mouse and keyboard inputs from your host into the VM?

In an ideal setup, imo, Botty should be running within the VM that way you can close it and let it do its own thing without overriding your host.

VladimirMakaev:
@Ezro WinAPI events from our keyboard & mouse can be detected that are not coming from physical device, which can theoretically flag activity for botting. By using VM you'd channel the input through the virtual drivers that would be detected by the game as actual input devices. So the game would have 0 way to detect if any mouse or keyboard automation is used. I'm actually interested in the use case myself as it's a good way to develop and test changes on multiple accounts simultaneusly.

brimmasch:
[...]

PR #563, opened 2022-02-18
Add ability to customize a window to search for with Win32 api

VladimirMakaev:
This is to address  #562 and allow running botty on the host with controlling a guest Hyper-V window.

For Hyper-V use configure in **custom.ini** (or params.ini)
```ini
[advanced_options]
hwnd_window_title={VMNAME} on .+
hwnd_window_process=
window_client_area_offset=0,51
```


mgleed:
I can't test this, can you confirm it works on vanilla and on VM?

VladimirMakaev:
If you can test that normal way works that'll help. I tested both on "my machine".  :)

mgleed:
Tested "normal" on my rig. Looks good

PR #568, opened 2022-02-19
Belt row detection

mischavandenburg:
Added a check that will detect how many rows the belt provides. 
Done by checking for the upper border of the belt in a certain ROI. 

I would love to get your feedback and hear if you see any problems with it. 

I have tested it with main.py by running new games, and I tested that it still drinks potions. 
Please let me know if there are other things I should test.


VladimirMakaev:
Potentially better solution would be to use InventoryCollection and count empty slots and known items being pots. @mgleed @Ezro are you doing anything with BeltManager in your UI work? Might wait for that to be merged instead

mischavandenburg:
> Potentially better solution would be to use InventoryCollection and count empty slots and known items being pots. @mgleed @Ezro are you doing anything with BeltManager in your UI work? Might wait for that to be merged instead

Thank you for your comment @VladimirMakaev , Gleed did mention that I would probably have to refactor after they are done with the overhaul and that is no problem for me.

mgleed:
> > Potentially better solution would be to use InventoryCollection and count empty slots and known items being pots. @mgleed @Ezro are you doing anything with BeltManager in your UI work? Might wait for that to be merged instead
> 
> Thank you for your comment @VladimirMakaev , Gleed did mention that I would probably have to refactor after they are done with the overhaul and that is no problem for me.

The overhaul is done now. I will be doing some other arrangements but we could probably fit this in. Perhaps me or @VladimirMakaev can help in the discord sometime to see how to leverage InventoryCollection to accomplish this. There's many ways to get belt height, but I think whatever way we can use that doesn't require creating a new asset and doesn't take any time to do would be optimal.

mischavandenburg:
> > > Potentially better solution would be to use InventoryCollection and count empty slots and known items being pots. @mgleed @Ezro are you doing anything with BeltManager in your UI work? Might wait for that to be merged instead
> 
> > 
> 
> > Thank you for your comment @VladimirMakaev , Gleed did mention that I would probably have to refactor after they are done with the overhaul and that is no problem for me.
> 
> 
> 
> The overhaul is done now. I will be doing some other arrangements but we could probably fit this in. Perhaps me or @VladimirMakaev can help in the discord sometime to see how to leverage InventoryCollection to accomplish this. There's many ways to get belt height, but I think whatever way we can use that doesn't require creating a new asset and doesn't take extra time away from the run would be optimal. 

Okay, thank you for the feedback. I will try to look into InventoryCollection and perhaps I would need some help. I am having a busy Week but I think I'll look at it in the weekend.

mgleed:
Sorry this didn't end up getting merged, I led you astray with poor timing! It sounds like you've gotten pretty busy but if you'd like to revisit this again lmk.

PR #569, opened 2022-02-20
Pot buying and stash maintenance during Arcane Run

D2RLegit:
Allows "run all corners" in params.
Pots and stash maintenance in between corner runs on arcane.


D2RLegit:
Re-opend one with new UI update

PR #570, opened 2022-02-21
Feature: OCR for item drops and item description boxes

mgleed:
Adds a parameter `use_ocr` under [advanced_options] in params.ini that will use OCR on item drops and item description box contents during stashing and print to debug log. This is preliminary PR to allow others to develop other features dependent on these functions suck as pickit filtering, checking key/tome quantities, etc. Enabling `use_ocr` is currently purely cosmetic and will increase pickit times somewhat, so default is turned off.


VladimirMakaev (approved):
you know what you're doing :)

PR #571, opened 2022-02-21
Bugfix: Suppress warning if no discord webhook set

mgleed:
Fixes warnings if webhook not set but discord messenger api is param. Unmasked by #569

### DIFF
--- a/src/messages/discord_embeds.py
+++ b/src/messages/discord_embeds.py
@@ -13,10 +13,10 @@ class DiscordEmbeds(GenericApi):
     def __init__(self):
         self._file = None
         self._psnURL = "https://i.psnprofiles.com/games/3bffee/trophies/"
-        try:
-            self._webhook = Webhook.from_url(Config().general['custom_message_hook'], adapter=RequestsWebhookAdapter(), )
-        except InvalidArgument:
-            if Config().general["custom_message_hook"]:
+        if Config().general["custom_message_hook"]:
+            try:
+                self._webhook = Webhook.from_url(Config().general['custom_message_hook'], adapter=RequestsWebhookAdapter(), )
+            except InvalidArgument:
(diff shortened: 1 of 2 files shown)


src/messages/discord_embeds.py
```diff
@@ -16,7 +16,8 @@ def __init__(self):
         try:
             self._webhook = Webhook.from_url(Config().general['custom_message_hook'], adapter=RequestsWebhookAdapter(), )
```
VladimirMakaev, line 17:
```suggestion
            hook = Config().general['custom_message_hook']
            if hook:
                 self._webhook = Webhook.from_url(hook, adapter=RequestsWebhookAdapter(), )
            else:
                 Logger.error("Nice message that custom_message_hook isn't set")

File: src/messages/discord_embeds.py [hunk 2]
```diff
@@ -16,7 +16,8 @@ def __init__(self):
...
         except InvalidArgument:
-            Logger.warning(f"Your custom_message_hook URL {Config().general['custom_message_hook']} is invalid, Discord updates will not be sent")
+            if Config().general["custom_message_hook"]:
```
VladimirMakaev, line 19:
```suggestion
              Logger.error("Message about error")
```

mgleed: @VladimirMakaev the trouble is it sends the error multiple times for some reason. I could put a debug line saying "not set" but then it'd spam it multiple times. I've rearranged the logic slightly in latest commit but otherwise I think it should be fine.

mgleed:
@VladimirMakaev the trouble is it sends the error multiple times for some reason. I could put a debug line saying "not set" but then it'd spam it multiple times. I've rearranged the logic slightly in latest commit but otherwise I think it should be fine.

egut125 (approved):

PR #576, opened 2022-02-22
Maintenance during arcane corners update with ui

D2RLegit:


D2RLegit:
Still working on making it able to run all corners from params work. Currently does not work with ID at cain turned on as it will fail runs when going to cain due to time out on dialogue screen.

D2RLegit:
With id at cain fix. Should be good to go.

mgleed:
This should probably be re-done with inspect_items() in mind. Do you want to modify this branch or start fresh? If you're still interested in salvaging we might be able to separate the town functions even more before fixing this so we don't have to be so redundant with code.

D2RLegit:
Closing, cant get to work good with ui overhaul for all corners. ill add inspect_items() though for maintenance to new pr

PR #582, opened 2022-02-23
Bugfix/enhancement: Function to show all ScreenObjects

mgleed:
Implemented a ui function that will return a list of all matching ScreenObjects on current screen.

Utilize in another new function in view.py `return_to_game()` that will return player to regular play state if other ui stuff are up like inventory, npc dialogue, etc.

Should solve issues with cain dialogue box occluding stash. 

Also implemented a fix for NPC quest dialogue disrupting the NPC routines. @D2RLegit 

Adjust a few assets and added some more ScreenObjects.


VladimirMakaev (approved):
Lgtm

PR #584, opened 2022-02-24
Bugfix: Properly dismiss NPC dialogue 

mgleed:
Thought this was fixed in #582 but it did not address all cases apparently--this method should be more robust. Tested and works nicely on NPC's that have quest dialogue, etc. but all you want to do is shop.


VladimirMakaev (approved):
lgtm

PR #586, opened 2022-02-24
Bugfix: Restore for loop in gamble

mgleed:


egut125 (approved):
What was the point of the loop? Attempt to transfer gold from certain tab to 0 tab a couples time just to make sure it all moves?

mgleed:
> What was the point of the loop? Attempt to transfer gold from certain tab to 0 tab a couples time just to make sure it all moves?

Yes. Need a smarter way of doing this eventually, but this will do for now.

PR #588, opened 2022-02-24
Feature / refactor: Town routine overhaul, scrolls/keys 

mgleed:
### Towning overhaul
-Now checks items every pickup and decides to drop, ID, sell, or stash at the start of on_maintenance(), so no longer will go to the stash needlessly
-Added a sell_junk char config, set to 0 if you want to drop all items that won't be kept, to 1 to sell all items that won't be kept
-When purchasing items, will use "shift" click to buy stacks if appropriate
-Consolidated certain inventory functions to transfer_items() function
-Reduced various redundant or excessively frequent actions with towning/shopping

### Item identification
-Now supports tome of ID in inventory and will utilize if present, otherwise will utilize cain if id_items config parameter = 1
-Will skip identification of items that have no filtered properties in pickit.ini; e.g., if you have `uniq_amulets=1` without any inclusion or exclusion properties it will pickup and keep but not identify. *If you have other items that need ID and you are using Cain, however, this may be overridden* 
-No item filtering with OCR yet implemented, but this is planned

### Consumables
-TP, ID, and key counts are now tracked with potion counts, utilizes OCR with open inventory 
-Will buy TP, ID, and keys from vendors as needed
-Smarter tracking of consumables during and after runs, should be more accurate now
-No longer sells/buysback tome of TP during repair

### Pickit
-Will now pickup TP scrolls and ID scrolls if these are needed and if you have these set to 1 in the pickit.ini config file.
-Keys, scrolls, and gold are lower priority during pickit sequence

### Gambling
-Simplified logic
-Now gambles a set number of times rather than attempting to completely drain stash

### Other
-Added color_match (color_filter) as a parameter to template_finder.

### Discord
-Will send picture of every kept drop as well as the read OCR text


src/inventory/common.py
```diff
@@ -35,4 +55,119 @@ def slot_has_item(slot_img: np.ndarray) -> bool:
...
+        # if selling, control+click to sell
+        # TODO: if purchasing, right-click to buy
+        # TODO: if purchasing stack, shift+right-click to buy
+        if (action == "drop" and not left_panel_open) or action in ["sell", "stash"]:
+            keyboard.send('ctrl', do_release=False)
+            wait(0.2, 0.4)
+        for item in filtered:
+            attempts = 0
+            while True:
+                mouse.move(*item.pos, randomize=4, delay_factor=[0.2, 0.4])
```
VladimirMakaev, line 142: Can extract a function or two here

VladimirMakaev (approved):
Looks solid overrall good job! I've tested a little bit, works fine. Didn't go to the corner cases though

src/inventory/common.py
mgleed: Thanks for looking things over! Really appreciate it.

I know that I *could* extract functions here, but I generally prefer to not extract things into separate functions if I'm sufficiently distant from the base scripts (game_controller, main, bot, etc.) and if I don't plan to use those functions anywhere else now or in the future, mostly to improve readability. I don't think the contents of this function in particular will need to be used elsewhere, so I'll probably leave them here.

VladimirMakaev: It's more about supporting this by another person some time later. When I see a good bunch of mouse & keyboard interactions in the same function I'm completely clueless what's going on and it's hard to reason what's going on without seeing it running. And it's located in the loop as well. A functions extracted in a form of "_hover_over_XX", "_drag_xxx_to_yyy", "_select_zzz" would boost readability of this code significantly without really any drawbacks

mgleed:
To do: 
Handle keys for act 2 drognan

PR #600, opened 2022-02-27
Add override_capabilities param to workaround problems with autodetection

VladimirMakaev:
For the folks having problems with autodetection and also the folks who want to use teleport charges without optimization


mgleed (approved):
Thank you!

PR #607, opened 2022-03-05
Support color filtering during template_finder.search()

mgleed:
Pass a color to template_finder.search() to filter both the template you are trying to match as well as the image you are searching. Easier and more accurate than having to add alpha for some cases.

Example:
```py
if (result := TemplateFinder().search(["YOU_HAVE_DIED", "NOT_ENOUGH_GOLD"], img, color_match=Config().colors["red"], use_grayscale=True, best_match=True)).valid:
        print(f"match: {result.score}")
```

Template images:
you_have_died
not_enough_gold

Filtered (and grayscaled) template:
result_20220305_113801_template

Images of interest:
result_20220305_113801_pre
result_20220305_113623_pre

Filtered (and grayscaled) images of interest:
result_20220305_113801_post
result_20220305_113623_post

Match on death image: 0.99

Match on out of gold image: 0.99

I know cthuzael has a PR involving some template / image filtering. I think his solution is decent but will need some time / refactoring to incorporate. This can be used right out the gate given we already use color_filter() all over the codebase.


mgleed:
Included in #588

PR #608, opened 2022-03-05
Improved eld attack sequence for Blizz sorc

D2RLegit:
added an extra attack sequences and added wait timers.


Steve6030 (commented):
Wait timer changes and extra sequence working well.

mgleed (approved):
Thank you

PR #625, opened 2022-03-10
Bugfix: Laggy inventory loading

mgleed:
On slow PCs or with high-latency NPC inventories will be slow to load and may be missed when starting `buy_consumables()`. Addressed this by implementing a function to determine whether an inventory is completely loaded. Tests for presence of tabs, blue slots, or red slots in respective ROI's.

Tested online/offline, seems to work fine.


src/inventory/vendor.py
```diff
@@ -140,4 +140,6 @@ def buy_item(template_name: str, quantity: int = 1, img: np.ndarray = None, shif
             return False
 
     Logger.error(f"buy_item: Desired item {template_name} not found")
-    return False
\ No newline at end of file
+    return False
+
+
```
VladimirMakaev, line 145: nit

VladimirMakaev (approved):

src/inventory/vendor.py
mgleed: I was just testing you ;)

PR #630, opened 2022-03-11
Added param to incrase lightness of grabbing image

MistrzHieroglif:
On some GPU (not supported officialy, in ex. integrated Intel GPU) grabbed image is too dark for recognizing any object. Added param to incrase lightness of grabbed image


mgleed:
If it's a problem on your personal machine, can't you just adjust the d2 in-game brightness?

PR #632, opened 2022-03-12
Remove: Codecov

aeon0:
Removing the codecov reporting since it is bound to my account and might cause some issues or at least is not nice to the eye to have a few red x on the PR :)
Leaving the coverage report generation just to keep that info available for anybody interested.


mgleed (approved):
<3

PR #641, opened 2022-03-15
Bugfix/refactor: Safer screen transitions for laggy setups

mgleed:
Some users with laggy setups struggling with save/exit and other routines. Working on implementing wait_until_visible and wait_until_hidden in various places of the code rather than hard-coded waits, etc.


src/char/necro.py
```diff
@@ -719,8 +716,7 @@ def to_trav(self):
                     wait(0.08, 0.15)
                     mouse.click(button="left")
                     Logger.debug("entering trav...")
-                    match = wait_until_visible(ScreenObjects.Loading, 2)
-                    if match.valid:
+                    if wait_until_visible(ScreenObjects.Loading, 2).valid:
```
mgleed, line 719: necro.py needs a lot of cleanup, but that can wait for another PR

PR #642, opened 2022-03-16
Bugfix: Some NPC's don't open to misc tab

mgleed:
Ormus occasionally opens to Weapons tab instead of Misc tab.

Implemented some tab functions in inventory/common.py. 

Will now select the 4th tab (index 3) whenever buying consumables.


VladimirMakaev (approved):
looks good

PR #647, opened 2022-03-18
added error handling for custom.ini when missing section header

Aimware0:
I had to record & slow down a video to see the error message since the cli closed so quickly due to a missing section header in the custom.ini. this fix it..


mgleed (approved):
Looks good, thank you

PR #654, opened 2022-03-20
Add beginnings of character skill detection

Ezro:


mgleed:
Marked stale. You're welcome to reopen if you'd like to continue work on this, ty Ezro

Ezro:
That's fair. I haven't looked at this since I created it, but the notion was that using this code we could auto-detect what keys are bound (and what the keys are).

The +1 of this would be that Botty could then automagically "play your character" without having to define which abilities to use, etc.

It's a bit lofty, but that was the initial idea.

PR #655, opened 2022-03-20
bugfix/anya_claws_best_match

alex-c-i:
Extended the functionality of TemplateFinder.search / TemplateFinder.search_and_wait by replacing the boolean best_find with 
a search_type param that can be one of 3 value: ALL_MATCH (This ones fixes the Anya shopper not looking through all the claws bug),
BEST_MATCH and FIRST_MATCH (this is the default just like before).

For BEST_MATCH and FIRST_MATCH the function returns a TemplateMatch just like before and for ALL_MATCH it now returns a list[TemplateMatch].

All calls with best_match True were replaced with search_type=SearchType.Best match.
All calls with best_match False had the param removed since FIRST_MATCH was the default anyway.

The Anya / Drognan shopper calls were replaced with ALL_MATCH to get all the claws / scepters.

Updated gitignore to include the IntelliJ generated files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,8 +50,3 @@ jobs:
         run: |
           C:\Miniconda\condabin\conda.bat activate botty
           coverage xml
-      - name: Upload Coverage to Codecov
-        uses: codecov/codecov-action@v2
-        with:
-          verbose: true
-          files: ./coverage.xml
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,8 @@ stats/
 htmlcov/
 coverage.xml
 utils/live-view/
+botty*.iml
+misc.xml
+vcs.xml
+modules.xml
+.idea/*
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,1 +1,0 @@
-comment: false
--- a/src/bot.py
+++ b/src/bot.py
@@ -15,7 +15,7 @@ from game_stats import GameStats
 from logger import Logger
 from config import Config
 from screen import grab
-from template_finder import TemplateFinder
+from template_finder import TemplateFinder, SearchType
 from char import IChar
 from item import ItemFinder
 from item.pickit import PickIt
--- a/src/bot.py
+++ b/src/bot.py
@@ -15,7 +15,7 @@ from game_stats import GameStats
 from logger import Logger
 from config import Config
 from screen import grab
-from template_finder import TemplateFinder, SearchType
+from template_finder import TemplateFinder
 from char import IChar
 from item import ItemFinder
 from item.pickit import PickIt
--- a/src/shop/anya.py
+++ b/src/shop/anya.py
@@ -147,11 +147,13 @@ class AnyaShopper:
                 # Search for claws
                 claw_pos = []
                 img = grab().copy()
+
                 claw_keys = ["CLAW1", "CLAW2", "CLAW3"]
-                for ck in claw_keys:
-                    template_match = TemplateFinder(True).search(ck, img, roi=self.roi_vendor)
+                template_matches = TemplateFinder(True).search_all(claw_keys, img, roi=self.roi_vendor)
+                for template_match in template_matches:
                     if template_match.valid:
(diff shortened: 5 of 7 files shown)


VladimirMakaev:
I'm personally not a big fan of return signature to be "X | list[X]" as it really hard to use it correctly as the client needs to know the implementation details and IDE can't suggest anything as well

I'd rather have a specialised separate method signature "search_all" that would return a list[X] instead.

alex-c-i:
As I'm new to Python I was also a bit unsure of x | list[x] and people on discord seemed to think its fine so I went ahead with it. 
Now that I think about it more it's definitely a valid point and it would be best to have 2 functions. 

Should I split the logic completely into 2 separate function ? Or is it acceptable if I rename the current function with a "_" prefix and delagate to it from 2 functions search (returns x) and search_all (returns list[x]) ? The intent being that outside callers should not call methods named _<name>, it seems to be a python idiom from what I can see even though nothing stops anyone from actually calling them aka there is no enforced private access in python.

VladimirMakaev:
You can probably have a private (_ prefixed) function that ALWAYS returns a list[TemplateMatch] and then structure the rest
via that private function.

search -> _search_all_internal(....). (in case of BEST MATCH picks the one that has a best match, in case of FIRST_MATCH the first one). _search_all_internal can also be made as generator so that in case of FIRST_MATCH can exit immediately after finding a match.
search_all-> _search_all_internal(...) (picks all matches)

I think keeping the enum is fine but probably makes sense to remove ALL_MATCH from there.

alex-c-i:
Yeah thats exactly what I was thinking, to have the private one always return a list.

Not sure how to keep the enum right now unless I'll have to declare another internal enum as well (since we dont want the callers to be able to pick ALL_MATCH using the normal search function). 
I will have to look at the code a bit and I'll figure out a clean way. I can probably work on this PR this week but not sure if I can in the next 2-3 days, most likely the weekend.

mgleed:
Thanks for working on this and appreciate @VladimirMakaev for the feedback. I didn't see a big issue with the Union output. I do agree though that the proposed alternatives are probably more clean. 

I think the optimal solution, however, should require no changes to the existing template_finder.search() calls in the codebase so we touch as few files as possible. But that just might be my bug-fixing exhaustion speaking. ;)

Excited to see what you come up with!

alex-c-i:
Not sure what happened to bot.py, the way i reverted was:

1. git checkout to the previous commit

2. create a new temp branch and restore only the template finder file and anya and dognan but with the correct calls to search_all

3. merge the temp branch into bugfix/anya_claws_best_match - my idea here was to have the merge being automatic and while it was it didnt do what I expected and update all the calls to search from all the files using the code in the temp branch (the one having the reverted code) but rather use the calls from the bugfix/anya_claws_best_match branch with the enum being used - not sure what the proper way to go in this case would have been, I though simply reverting the last commit would not be OK but this only got me into more of a mess.

For some reason only bot.py had the call with the enum so I reverted that one by hand afterwords.

mgleed:
> Not sure what happened to bot.py, the way i reverted was:
> 
> 1. git checkout to the previous commit
> 2. create a new temp branch and restore only the template finder file and anya and dognan but with the correct calls to search_all
> 3. merge the temp branch into bugfix/anya_claws_best_match - my idea here was to have the merge being automatic and while it was it didnt do what I expected and update all the calls to search from all the files using the code in the temp branch (the one having the reverted code) but rather use the calls from the bugfix/anya_claws_best_match branch with the enum being used - not sure what the proper way to go in this case would have been, I though simply reverting the last commit would not be OK but this only got me into more of a mess.
> 
> For some reason only bot.py still had the call with the enum being used so I reverted that one by hand afterwords.

Thanks again for your work on this. If you need to just open a new PR or something on a new branch and just copy/paste the relevant code that might help with some of the merge headaches.

le3bl:
I implemented a similar feature in my PR (676) that I am working on adding more shopping/shoppers to.
https://github.com/aeon0/botty/pull/676

alex-c-i:
Closed and opened #683 .

PR #662, opened 2022-03-27
Leveraging OCR to read exp and track in game_stats

egut125:
Start of each game and end of each run OCR exp
game_stats stores starting EXP and current EXP
game_stats summary calculates XP gained, XP per game, XP per hour

### DIFF
--- a/src/ui/player_bar.py
+++ b/src/ui/player_bar.py
@@ -14,6 +14,8 @@ def get_experience():
     wait(0.05)
     # crop roi
     img = grab()
+    
+    mouse.move(x_m, y_m-50, randomize = (8,1))
     crop = cut_roi(img, Config().ui_roi["xp_bar_text"])
     ocr_result = Ocr().image_to_text(
         images = crop,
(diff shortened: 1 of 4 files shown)


src/ui/player_bar.py
```diff
@@ -14,6 +14,8 @@ def get_experience():
     wait(0.05)
     # crop roi
     img = grab()
+    
+    mouse.move(x_m, y_m-50, randomize = (8,1))
```
mgleed, line 18: I don't believe grab() includes the mouse, does it? Never seen the mouse in any screenshots from cv

egut125: Added this so that the mouse goes back into the main area as the next action Botty does is try and tp back to town. So figured would just put the mouse back into the game area. 

Otherwise Botty would try and right click tp,  but mouse is over xp bar and tp isn't made, Botty waits until the timeout and repositions to tp again. 

Other thought was have the tp process make sure the mouse is on screen somewhere clickable but this was easier haha

PR #676, opened 2022-03-31
Feature/enhancement Add more shoppers and better shopping

le3bl:
I intend on creating several shoppers ahead of ladder and I wanted to start by setting it up so they can be added dynamically where there will not be a need to modify shopper.py each time a new shopper is added/removed.


mgleed:
I like this PR a lot, this was a ton of work! Sorry it's taken so long to review. One concern I have is that we are planning to merge #724 shortly which will get rid of template matching for item properties, so all the new templates/assets that would be added in this PR would have to be dealt with after merging. Another concern I have is template_finder.py the new multi_search() function duplicates too much code. I'd like to see an implementation more like #683 for that.

mgleed:
Closing for now. Will need to revamp shopper bot after #724 at which point we can revisit this. Really like what you did here.

PR #679, opened 2022-04-01
Quick list manipulation to remove anything before experience in the list

egut125:
Numerous xp ocr issues from ocr finding some random character before the experience word. This will ensure those are dropped from the list and it starts with experience as expected.


mgleed (approved):
Still dies on level 99, so I added a conditional statement. It doesn't even seem to calculate the level correctly for a level 99 (shows level 1) but otherwise works for now. :) 

I also put in a try/except since this is mostly a cosmetic feature anyway.

PR #683, opened 2022-04-01
Extended the functionality of TemplateFinder by adding an internal _s…

alex-c-i:
…earch_all_internal function that returns a list and takes a _SearchType param:

- FIRST_MATCH: returns after a match is found (default); uses the current implementation in the search function
- BEST_MATCH: returns the best out of all matches; uses the current implementation in the search function
- ALL_MATCH: returns all the matches; implemented by removing the current result after each match

Replaced the implementation of the search function with an appropriate call to _search_all_internal.

Added a new function search_all that delegates its work to _search_all_internal.

The Anya / Drognan shopper search calls were replaced with search_all calls.

Updated gitignore to include the IntelliJ generated files.

### DIFF
--- a/src/template_finder.py
+++ b/src/template_finder.py
@@ -10,8 +10,15 @@ import time
 import os
 from config import Config
 from utils.misc import load_template, list_files_in_folder, alpha_to_mask, roi_center, color_filter
+from enum import Enum, auto
 
 template_finder_lock = threading.Lock()
+
+class _SearchType(Enum):
+    FIRST_MATCH = auto()
+    BEST_MATCH = auto()
+    ALL_MATCH = auto()
+
 
 @dataclass
 class TemplateMatch:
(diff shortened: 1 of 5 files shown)


src/template_finder.py
```diff
@@ -89,6 +96,104 @@ def search(
...
+        :param ref: Either key of a already loaded template, list of such keys, or a image which is used as template
+        :param inp_img: Image in which the template will be searched
+        :param threshold: Threshold which determines if a template is found or not
+        :param roi: Region of Interest of the inp_img to restrict search area. Format [left, top, width, height]
+        :param normalize_monitor: If True will return positions in monitor coordinates. Otherwise in coordinates of the input image.
+        :param use_grayscale: Use grayscale template matching for speed up
+        :param color_match: Pass a color to be used by misc.color_filter to filter both image of interest and template image (format Config().colors["color"])
+        :return: Returns a list of TemplateMatch objects with a valid flag
+        """
+        return self._search_all_internal(ref, inp_img, threshold, roi, normalize_monitor, _SearchType.ALL_MATCH, use_grayscale, color_match)
```
aeon0, line 172: Whats to point of having an private (internal) function that is just mapped from a public one?

alex-c-i: 1. Not duplicating code
2. Cleaner API by having:
a. a public seach by first / best match function that retuns a TemplateMatch 
b. another search_all function that returns a list of TemplateMatch

mgleed (changes requested):
Fails tests from #753

src/template_finder.py
```diff
@@ -159,86 +265,64 @@ def search(
...
+                    min_val, max_val, min_loc, max_loc = cv2.minMaxLoc(res)
+
+                    if self._save_last_res:
+                        with template_finder_lock:
+                            self.last_res = deepcopy(res)
+
+                    if max_val > threshold:
+
+                        res[max(max_loc[1]-h//2, 0):max_loc[1]+h//2+1, max(max_loc[0]-w//2, 0):max_loc[0]+w//2+1] = 0# eliminate current result from results
+
```
mgleed, line 288: You could use utils.misc.mask_by_roi() here if you'd like.

mgleed:
Closed by #781 , thanks again for your help with this!

PR #687, opened 2022-04-03
Condsensed discord message will be the only option

D2RLegit:
Condensed message will be default. removed the option from params.

### DIFF
--- a/src/messages/discord_embeds.py
+++ b/src/messages/discord_embeds.py
@@ -71,7 +71,6 @@ class DiscordEmbeds(GenericApi):
     def send_message(self, msg: str):
         msg = f"{Config().general['name']}: {msg}"
         e = discord.Embed(title=f"Update:", description=f"```{msg}```", color=Color.dark_teal())
-        e.set_thumbnail(url=f"{self._psnURL}36L4a4994.png")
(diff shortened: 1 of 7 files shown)


src/messages/discord_embeds.py
```diff
@@ -71,8 +71,7 @@ def send_gold(self):
     def send_message(self, msg: str):
         msg = f"{Config().general['name']}: {msg}"
         e = discord.Embed(title=f"Update:", description=f"```{msg}```", color=Color.dark_teal())
-        if not Config().general['discord_status_condensed']:
-            e.set_thumbnail(url=f"{self._psnURL}36L4a4994.png")
+        e.set_thumbnail(url=f"{self._psnURL}36L4a4994.png")
```
egut125, line 74: Don't want the image in the condensed version.  It was only adding the image if it was not condensed

Remove this line


```suggestion

```

egut125 (changes requested):
Need to remove from readme.md as well.

src/messages/discord_embeds.py
D2RLegit: fixed

egut125 (approved):

PR #689, opened 2022-04-03
(WIP) Added hooks + some webserver stuff and some database stuff

Aimware0:
The idea behind hooks is separate the core of the bot while being able to create extensions, as you will notice while reviewing the /scripts folder (all the python files inside scripts are automatically executed. the /scripts/utils is not automatically executed). And there is some database / website wip code too


src/bot_events.py
```diff
@@ -0,0 +1,92 @@
...
+    # be nice to other developers and make sure you put it in here so they know it exists
+    # thanks
+    
+    # bot hooks
+    "on_bot_init": {},
+
+    "on_create_game": {},
+    "on_end_game": {},
+
+    "on_run_pindle": {},
```
ramfjord, line 15: Consider that the interface for bots effectively allow you to implement a singular callback function for each of these "on_run_foo"s at the moment.  This changeset is basically just letting you define multiple callbacks for when you enter a pindle run.  Is there actually a reason you would ever want to do that? In addition, when you register an `on_run_pindle` callback, it's may be less clear what arguments your callback function will receive than it previously was when you are implementing an abstract method.

src/bot_events.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # websockets / website hooks
+    "recv_website_request": {},
+}
+
+
+"""
+    This is a bit unorthodox, but the way this works is you do
+
+    hook.Call("on_pickup_item", item_name=item, location=location, ocr_text=ocr_text) you would put this line where the bot picks up an item
```
ramfjord, line 47: You could consider defining *this* interface alongside the name of a callback, possibly alongside type info, and basically validating in `hook.Call` that all the right parameters are passed in or raising a valueError, to make both debugging and using these callbacks easier

ramfjord (commented):
Do all these callbacks need to share this interface?  It might make sense for certain hooks to only allow one implementation, which the simple `on_foo` methods already enforce in the base bot. I think the fact that they'll be invoked from separate threads, and potentially modify the same global UI state is an issue. I didn't really look at the whole websocket thing that seems to be more WIP, but that also seems fairly separate to the callback code, except that it uses callbacks to write to the server.  This would potentially be easier to review though if you split out the callback implementation, and then had another PR for the DB/Server stuff.

Aimware0:
I've replied messaged to the user above on discord and we discussed it.

PR #700, opened 2022-04-13
Imagine: being able to `from health_manager import mp, hp` in your bot!

ramfjord:
This PR makes the health_manager effectively into a singleton, which maintains a single global state, rather than keeping it as a class that can be instantiated multiple times.  Unless there is a use case for having multiple health managers running simultaneously in the same process, there's no reason to allow it to instantiate multiple versions of its state.  There are multiple patterns in python to do this, I could have decided on an actual singleton class decorator implementation, or used the "borg" pattern to set the state of all health_managers the same, but this seemed by far the most idiomatic and least verbose.  I've attempted to somewhat limit the changes here, though I've got tabs configured for this project in my editor from editing the ranged bot and other files - so... LMK if you want those back as spaces

The reason I wanted this in the first place was because I'd love to be able to validate that a spell was cast from within my bot by ensuring the mana went down.  There currently seems to be no way to cleanly check the HP/MP from within a bot.  This will allow any bot to simply `from health_manager import hp, mp` to call `hp()` and `mp()` from within their bot class.  Though... y'all might be thinking about my logic for changing this from being a class (no two instantiated in a single process), and realizing the same holds true for a bot, I don't see any reason to prioritize refactoring that, and inheritance can be a useful tool for bot creation.

I've only copied in files that are working for me, still need to test this more thoroughly.

### DIFF
--- a/src/health_manager.py
+++ b/src/health_manager.py
@@ -2,9 +2,10 @@ from inventory import belt
 from pather import Location
 import cv2
 import time
+import threading
 import keyboard
 from utils.custom_mouse import mouse
-from utils.misc import wait
+from utils.misc import wait, time_since
 from logger import Logger
 from screen import grab
 import time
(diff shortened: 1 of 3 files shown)


src/health_manager.py
```diff
@@ -13,140 +14,179 @@
 from ui import view, meters
 from ui_manager import ScreenObjects, is_visible
 
-pause_state = True
+_pause_state = True
```
ramfjord, line 17: given that we're exposing getters and setters, giving this variable an `_` prefix will indicate it's "private" status

src/health_manager.py
```diff
@@ -13,140 +14,179 @@
...
 
 def get_pause_state():
-    global pause_state
```
ramfjord, line 19: The global keyword is only necessary when modifying variables, not when you're just accessing them in a function

mgleed (commented):
This looks good to me, I might request another reviewer or two with more threading exp to help review. Only nitpicky thing I'd request adjusting is the tab spacing convention--yours appears to be wider than what we use throughout the project.

mgleed:
@Ezro will you take a look at this?

mgleed:
@ramfjord Hey, I'm trying to catch up on PR's. I see @aeon0 left a review 25 days ago. What is left to be done here?

Ezro:
> Considering that we already have many global / singelton / functional stuff by now and it seems to work okay and we will probably never do any proper unittests, I'd be fine with it. The implementation is also fine.
> 
> The only thing I am a bit sceptical is updating hp and mp every run when health manager is running. What if it is not running e.g. in town. Now you might get different values from health manager vs meters. I'd stick with just doing this when you need the health:
> 
> ```python
> from ui import meters
> meters.get_health()
> ```
> 
> Apart from that, please 4 spaces indent.

Historically, I've actually ran into issues stemming from this: the bot would kill pindle, think it left the game (thereby turning off the health_manager, I think?) and then the zombies rise and kill the bot (since the bot no longer was detecting if it needed to potion or chicken).

Not sure if this is still an issue, and it doesn't feel that relevant to this PR, but I figured I'd chime in. lol

mgleed:
@ramfjord Thanks for this PR. There were changes requested over a month ago but haven't seen any updates. I think this could still end up being a good feature, but closing PR for now as it's stale. If you'd like to finalize it we can certaintly reopen it when you're ready.

PR #702, opened 2022-04-15
BugFix: Allow 2 lines of tooltip

roozhou:
Fix item_cropper failure on D2R 2.4


Aimware0 (approved):
tested and works well gj and ty.

PR #703, opened 2022-04-15
Bugfix: False empty stash tabs due to item tooltip

mgleed:
Fixes a bug introduced in 7.2 after attempting to fix another bug. 

The current bug manifests with users having their botty quit after attempting to stash/sell/drop an item whose tooltip was too large causing calc_roi() to fail. Calc_roi will still fail with this bugfix, but now anytime an item is to be transferred it will first make sure that inventory slot is not empty. If it is empty, that item will be skipped.

Here is an example of Herald of Zakarum stashing and not causing game to quit when it previously did.

image

While fixing calc_roi would be nice, this should hopefully no longer be a necessary function with implementation of HoveredItem from d2r_image library by @Ezro, hoping to have 0.8.0 out soon.

### DIFF
--- a/src/inventory/common.py
+++ b/src/inventory/common.py
@@ -10,24 +10,9 @@ from utils.misc import wait, trim_black, color_filter, cut_roi
 from inventory import consumables
 from ui import view
 from screen import convert_screen_to_monitor, grab
-from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


aeon0 (approved):
Hacky but should do it for now

PR #706, opened 2022-04-15
fix: Remove ;

aeon0:
Is this a newly introduced operator in the latest python versions?


mgleed (approved):
Don't think so. :) @VladimirMakaev fyi

PR #708, opened 2022-04-17
BugFix: Fix indicator_location_to_tab_count

roozhou:
1) 4 tabs instead of 3
2) Return value should be integer when failed


aeon0 (approved):
Good catch, the False == 0 will return True 🤦

PR #709, opened 2022-04-17
feat/restricted-mouse-clicks

aeon0:
Since the beginning of botty there were few reports that equipped items or charms were lost. In all of these cases, the reason is most likely some config errors. 
But just to make sure this does not happen check before every click if inventory is open and restrict clicking to not click in equipped area and in the restricted inventory area (depending on `num_loot_columns`)


src/config.py
```diff
@@ -392,13 +392,6 @@ def load_data(self):
         self.ui_roi = {}
         for key in self.configs["game"]["parser"]["ui_roi"]:
             self.ui_roi[key] = np.array([int(x) for x in self._select_val("ui_roi", key).split(",")])
-        # calc roi for restricted inventory area
-        self.ui_roi["restricted_inventory_area"] = [
-            int(self.ui_pos["inventory_top_left_slot_x"] + self.ui_pos["slot_width"] * self.char["num_loot_columns"]), # left
-            self.ui_pos["inventory_top_left_slot_y"], # top
-            int(self.ui_pos["slot_width"] * (10 - self.char["num_loot_columns"])), # width
-            int(self.ui_pos["slot_height"] * 4) # height
-        ]
```
mgleed, line 395: @aeon0 if you still want this to be in Config() I think that'd be fine! I'd just also consider having another calculated attribute like `unrestricted_inventory_area` to capture the region of the other part of inventory. I'd also then update anywhere `personal.specific_inventory_roi()` is used.

mgleed:
Well, I tried to replace the TemplateFinder method with ScreenObjects but unfortunately ran into circular imports I didn't want to resolve. Also ran into the same trying to use inventory/personal.py's function for determining reserved/restricted inventory but also ran into circular import. Ended up extending your changes in config.py and modified the codebase to utilize it.

custom_mouse.py tests work, let me know what you think. @aeon0

aeon0:
I think that circular import hints that sth isn't right with how we structured the code right now. 
ScreenObject should probably be in its own file and the is_visible and detect_screen_object are probably better off in the TemplateFinder()?

mgleed:
> I think that circular import hints that sth isn't right with how we structured the code right now. ScreenObject should probably be in its own file and the is_visible and detect_screen_object are probably better off in the TemplateFinder()? But we can merge this now and do that in a different PR.
> 
> Putting the rois in the config like this is good imho.

I think you're right. Part of the issue is mouse is imported in pretty much every file. :P Open to ideas for sure.

PR #710, opened 2022-04-17
Updated pot drinking logic

Aimware0:
Now limit the drinking of the HP / MP pots to every 10 seconds (since this is how long they last). or if the localplayer / merc reaches 100% HP / MP the timer will reset 0 seconds, 3 separate timers relative to HP / MP / MERC HP are tracked

### DIFF
--- a/src/health_manager.py
+++ b/src/health_manager.py
@@ -79,10 +79,6 @@ class HealthManager:
         self._did_chicken = False
         start = time.time()
 
-        lp_hp_potion_delay = uniform(9, 10)
-        lp_mp_potion_delay = uniform(9, 10)
-        merc_hp_potion_delay = uniform(9, 10)
-
         while self._do_monitor:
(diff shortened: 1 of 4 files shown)


mgleed (approved):

src/health_manager.py
```diff
@@ -85,40 +87,46 @@ def start_monitor(self):
             if is_visible(ScreenObjects.InGame, img):
                 health_percentage = meters.get_health(img)
                 mana_percentage = meters.get_mana(img)
+
+                # reset delay if at any point a meter becomes full
+                lp_hp_potion_delay = 0 if health_percentage == 1 else uniform(9, 10)
+                lp_mp_potion_delay = 0 if mana_percentage == 1 else uniform(9, 10)
```
aeon0, line 93: Did you check if health and mana percentage actually hit 1? And not maybe 0.99?

PR #711, opened 2022-04-17
Fix: Increase item drop padding from 1 to 2 spaces, pad gold

mgleed:
Increased padding to two spaces on each side of item drop rather than 1 space on each side. Also added space to right side of gold (can't pre-pad the gold integer afaik)

should hopefully address cases like the two below.

image
image


aeon0 (approved):
sure

PR #714, opened 2022-04-19
Organized Nip pickit for OCR

D2RLegit:


aeon0 (changes requested):
- Lets not have whitespaces in any folder names or file names
- Lets keep the convention of lowercase names only. Then we never have to worry about case sensitivity.
- Dont we want to have the files as .nip?

D2RLegit:
Not going this route

PR #724, opened 2022-04-21
Major feature: NTIP pickit syntax integration, OCR pickit, OCR item property filtering

mgleed:
* Add first pass at integration

* Add Legit's nip files as base nip files

* Replace existing image_to_text calls with d2r-image ocr

* Trim existing Botty OCR implementation

* Completely remove existing ocr.py in lieu of d2r-image

* remove tesserocr .whl and redundant dependencies

* remove tessdata

* working on removing old logic, reworked ItemText model, etc.

* fix .data -> .img

* update item_cropper _main_

* fix res result

* start integrating ground loot

* update item finder test

* print drop info

* everything is semi-working it appears

* improved pickit a bit

* Added more nip stuff

* optimized & cleaned pickit code

* fixed error due to merge

* Fixed some bugs relating to IDing and inf loop picking up

* added nip source files

* updated nip imports

* Fixed type with rejuv pot

* Added further clarification in lexer of what stage it is at

* Added further clarification in lexer of what stage it is at

* Added support for gold pickup

* Avoid highlighting & timeout during pickit

* Added temp bugfix for gold

* Big code clean up 95% done with pickit.py

* Added delay for checking inventory full

* Cleaned up imports

* Moved move_cursor_to_hud call to correct place

* Better disable functionality for nip files

* Added support for new nip operator '[idname]'

* Removed unneeded file

* duct-taped an error

* Made [idname] actually work

* Added file needed for [idname] intergeration

* Added char.capabilities.can_teleport_with_charges fail pickup

* update pickit

* Switched tokkens from being generator to list

* Fixed [type] never actually working correctly..

* Cleaned up & and get ready for nip

Co-authored-by: Ezro <justinmvalentin@gmail.com>
Co-authored-by: Your Name <you@example.com>
Co-authored-by: Riddle <mrthrowaway30293@gmail.com>


src/d2r_image/ocr.py
```diff
@@ -0,0 +1,255 @@
...
+    ocr_confidence_threshold: float = 0.9
+    ) -> str:
+    confidences = [x/100 for x in confidences]
+    if all(x >= ocr_confidence_threshold for x in confidences):
+        return original_text
+    #print(f"{original_text} {confidences}")
+    text = original_text.replace('\n', ' NEWLINEHERE ')
+    corrected_text = ""
+    word_count = 0
+    for word in text.split(' '):
```
ramfjord, line 226:
```
for word in text.split(' '):
  word = word.strip()
```


==

```
for word in map(str.strip, text.split(' ')):
```

Just a style preference though

src/d2r_image/ocr.py
```diff
@@ -0,0 +1,255 @@
...
+        word = word.strip()
+        if word and word != "NEWLINEHERE":
+            try:
+                #print(confidences[word_count])
+                if confidences[word_count] <= ocr_confidence_threshold:
```
ramfjord, line 231: It might actually be good to filter on whether or not the word is already *in* the word list before trying to fuzzy find it.  In fact, it might be worth going to the fuzzy finder even with a high ocr confidence, if the word isn't one that we recognize.

src/d2r_image/ocr.py
```diff
@@ -0,0 +1,255 @@
...
+    confidences = [x/100 for x in confidences]
+    if all(x >= ocr_confidence_threshold for x in confidences):
+        return original_text
+    #print(f"{original_text} {confidences}")
+    text = original_text.replace('\n', ' NEWLINEHERE ')
+    corrected_text = ""
+    word_count = 0
+    for word in text.split(' '):
+        word = word.strip()
+        if word and word != "NEWLINEHERE":
```
ramfjord, line 228: Another minor stylistic preference, feel free to dismiss this from me: I kind of prefer to put the shorter end of an if statement at the top in a method like this.  When I see this, I'm immediately thinking "ok, so we need to handle word is None, word is False, and word == "NEWLINEHERE" as special cases possibly", but then I see at the bottom that we're basically skipping this if word is None? I would prefer to explicity handle the single cases here like:

```python
if word is None:
  next # This seems to be the current behavior
elif word == "NEWLINEHERE"
  corrected_text += "\n"
else:
  <main method>
```

In this case, it's actually seems likely to me that there's no way `word` could actually be `None` or `False`, or even `''`, so I think it would be safe to drop the top part, but the general goal of these [guard clause](https://deviq.com/design-patterns/guard-clause) style if statements is to avoid nesting if clauses unecessarily by putting the short ones that can `return`, `break` or `next` at the top, and trying to keep most of it going through the general case.  But I can also see why you might have wanted to put the meat of the function at the top and handle the exception at the bottom.

src/d2r_image/ocr.py
```diff
@@ -0,0 +1,255 @@
...
+            try:
+                #print(confidences[word_count])
+                if confidences[word_count] <= ocr_confidence_threshold:
+                    alphanumeric = re.sub(r"[^a-zA-Z0-9]", "", word)
```
ramfjord, line 232: I thought that `sub` might only do the first occurrence, but it definitely does them all 👍 .  BTW, `r"\W"` is an exiting character class for non-alphnumeric characters

src/d2r_image/ocr.py
ramfjord, line 226: Also re. splitting in general, and I'm not sure if you would actually want to do anything about this: it's kind of a shame that we can't take advantage of knowing the sort of phrases the game is likely to throw at us, aka the context.  If I as a player see "C_______ VOUGE" I have a pretty good idea of what that first word might be, even if I thought I saw a bunch of 5s instead of S's.  This is just something I was thinking about though, I think this way of doing things with the right settings seems like it could work totally fine without the context of where a word is.  Passing in some context could also help make `word_list` smaller in a way - even if you only knew the word before or after, it could drastically limit the candidate words.  If you could quickly look up all of the words that come before `wand` in various game items, for instance.  But yeah, probably this speculation I'm giving you is completely unnecessary and premature optimization is the root and all that

src/d2r_image/ocr.py
mgleed: Part of this is that I'm confident I have like... 99.9% of the strings in the wordlist. If OCR has a high confidence on a word then I suspect it's probably a true word. Had to add a bunch of words to the wordlist tonight because I've chosen now to correct words that have punctuation attached like `Cost:` `Value:`. By not checking every word i give a little wiggle room for cases I might not have anticipated.

src/d2r_image/ocr.py
mgleed: Yes, I like this suggestion. I think you're right that it won't happen, but I still kept `if not word: continue`. I can't remember if when I made this function if I ran into some weird case and had to put it there. But yes otherwise making changes as suggested.

src/d2r_image/ocr.py
mgleed: Ended up changing how this piece works a bit. Now only checking to see if there are alphabetic characters within a string.

hutber:
Oh boy!!! Its a huge PR lol but I am very much looking forward to this one :D

NeoIsrafil:
I'm sooooo looking forward to this being running fully.  I pulled it, and the bot seems to have a very hard time reading either the tomes, or the nodes on the map... at least on today's version, but the new way of writing the pickit is SO AMAZING... its gonna be incredible.  <3  guys, y'all rock, looking forward to playing with it!

eveDude:
Would also be willing to test it out.
Do we have maybe a list of stuff to test for? :)

PR #725, opened 2022-04-21
Maintence During arcane run

D2RLegit:
Checks for pots/stashing/item selling inbetween arcane corner runs.


mgleed (changes requested):
Have you tested this? All inspect_items will do is return a list of items or drop them. Would have to modify inspect_items() for people who have sell_items=1 in params. All we could do with minimal effort now is have botty drop stuff in the field or in town. Stashing or selling would take considerable effort or duplicate code.

PR #727, opened 2022-04-22
Bone necro

slapchop58:
Added an implementation for bone_necro with support for all runs except Chaos. 

Added cast_in_arc to i_char.py, which is heavily used by the bone_necro implementation. This function allows a skill to be cast more realistically and with reduced downtime.  The mouse is held down for some duration, and moved around while held in order to continue casting randomly within an arc (degrees is a parameter).  

This means that we cast at max speed as opposed to many other implementation in this project where there are additional delays between casts - Usually a cast, then a delay, mouse movement (with it's own delay), another delay, then another cast. 

The only downside to this approach previously, was that if the cast started on an enemy and that enemy died, the skill would stop casting for the rest of the duration. 2.4 removed this downside and the spell will continue to cast.

The existing cast times are tailored for a relatively high level and well geared necro. Testing was on a level 90 char with nearly fully synergized bone spells at level 39 and 125 FCR. To adjust for a less or better geared character, you can set the damage_scaling option under [bone_necro]. ie: damage_scaling=2 will cast all skills twice as long.

### DIFF
--- a/config/params.ini
+++ b/config/params.ini
@@ -222,6 +222,16 @@ corpse_explosion=
 clear_pindle_pack=
 heart_of_wolverine=
 
+[bone_necro]
+teleport=
+teeth=
+bone_armor=
+clay_golem=
+bone_wall=
+bone_spear=
+bone_spirit=
+damage_scaling=1
+
 ; ===========================
 ; ==== Builds: Basic ====
 ; ===========================
--- a/config/params.ini
+++ b/config/params.ini
@@ -230,6 +230,8 @@ clay_golem=
 bone_wall=
 bone_spear=
 bone_spirit=
+; Damage scaling. One setting to adjust kill speed to account for better/worse gear. 
+; ex: 2 will cast all attack skills twice a long, .5 half as long. Currently only used by bone necro
 damage_scaling=1
 
 ; ===========================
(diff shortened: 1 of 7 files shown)


Aimware0:
I would love to see a video of this if you have one or can make one, seems like good work

slapchop58:
Sure. Here you go: https://youtu.be/VhaMYoFtcQU

Aimware0:
I like it, the only thing I would say is to refrain from using ascii color codes inside Logger methods. There was actually just a PR to fix the logger because it wrote color codes inside the info.log folder. the video looks good :D

config/params.ini
```diff
@@ -222,6 +222,16 @@ corpse_explosion=
...
 
+[bone_necro]
+teleport=
+teeth=
+bone_armor=
+clay_golem=
+bone_wall=
+bone_spear=
+bone_spirit=
+damage_scaling=1
```
Aimware0, line 233: also should probably add a comment here relative to what you said in your PR for the users who don't look at PRs. "you can set the damage_scaling option under [bone_necro]. ie: damage_scaling=2 will cast all skills twice as long."

mgleed: Rather than comment the config (technically you can do both I guess), will you update Readme.md? Otherwise lgtm. @slapchop58

jhumble: Sure. I can take care of both those. Those logger comments just got copied from the poison necro implementation that I copied over and modified as I was building this out.

mgleed (approved):
Thank you for addressing those changes. Sorry for the delay.

PR #730, opened 2022-04-22
New Hydra Build

egut125:
Only runs pindle shenk eld so far

### DIFF
--- a/config/params.ini
+++ b/config/params.ini
@@ -141,6 +141,12 @@ blizzard=
 ; nova must be right skill, hotkey required
 nova=f2
 
+[hydra_sorc]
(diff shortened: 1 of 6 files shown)


config/params.ini
```diff
@@ -141,6 +141,12 @@ blizzard=
 ; nova must be right skill, hotkey required
 nova=f2
 
+[hydra_sorc]
```
mgleed, line 144:
```suggestion
[hydra_sorc]
; only supports Pindle and Eldritch currently
```

mgleed (changes requested):
Looks good to me. If you'll just document somewhere like in params.ini or something that it only supports Eldritch and Pindle that'll be helpful for support. :)

mgleed (approved):

PR #733, opened 2022-04-23
BugFix: Do not write ANSI color codes to info.log

roozhou:
Most text editors do not like ANSI color codes


mgleed (approved):
Thank you!!

PR #734, opened 2022-04-23
feat/cached-img

aeon0:
I checked it and the cache is used quite frequently as we have many parallel things that call the screen grab(). Only downside is that we might have a maximum delay of 20ms to get the "new" frame.

TODO: making it thread safe...

### DIFF
--- a/src/pather.py
+++ b/src/pather.py
@@ -641,7 +641,7 @@ class Pather:
                 img = grab(force_new=True)
                 # Handle timeout
                 if (time.time() - last_move) > timeout:
-                    if is_visible(ScreenObjects.WaypointLabel):
+                    if is_visible(ScreenObjects.WaypointLabel, img):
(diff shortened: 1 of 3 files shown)


mgleed:
I love this idea.

I'm curious--if we know a frame lasts 40 ms why do we check if within 20 ms?

Gonna tag a few people I think this might be interesting to. @Ezro @Aimware0 @ramfjord

aeon0:
I was thinking nyquist theorem, but actually the "wavelength" is 80ms. Thus 40ms will be good. Thx for pointing this out.

mgleed (approved):
Thanks for doing this. Is this working? Any reason not to merge?

aeon0:
I did some testing and it felt like I saw some issues where we would click e.g. on waypoints with a few frames delay or over teleporting. But could have been just random bug. I will test this a bit more and merge it myself once I am more convinced that it is fine.

PR #736, opened 2022-04-26
Transmute All Gems

badbrad2889:
I added the ability to transmute all gems and select which kind of gems you would like to transmute in the params.ini file


D2RLegit:
Can't we make it as "transmute gems" as one line instead of having them all seperate in params?

TallenKerl:
> Can't we make it as "transmute gems" as one line instead of having them all seperate in params?

That's not a good idea because some people keep specific gems for upgrading specific runes or (in my case) rerolling certain items with chips is more effective than with perfect

badbrad2889:
@D2RLegit As TallenKerl said I built it this way so the user has the choice which gems they want to keep.

config/params.ini
```diff
@@ -111,6 +111,10 @@ gamble_items=
 stash_destination=3,2,1,0
 ;how often we want to run transmute routine(e.g. every 100 games)
 transmute_every_x_game=20
+transmute_chipped=1
+transmute_flawed=1
+transmute_std=1
+transmute_flawless=1
```
VladimirMakaev, line 117:
```suggestion
transmute="chipped,flawed,standard,flawless" ; 
```

VladimirMakaev (changes requested):

src/transmute/transmute.py
```diff
@@ -17,6 +17,36 @@
...
+    "INVENTORY_TOPAZ_FLAWED",
+    "INVENTORY_AMETHYST_FLAWED",
+    "INVENTORY_SAPPHIRE_FLAWED",
+    "INVENTORY_DIAMOND_FLAWED",
+    "INVENTORY_RUBY_FLAWED",
+    "INVENTORY_EMERALD_FLAWED",
+    "INVENTORY_SKULL_FLAWED"
+]
+
+STD_GEMS = [
```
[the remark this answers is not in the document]
VladimirMakaev: Please use the word "standard" everywhere since we use full names for all other qualities. So people like me with OCD don't suffer too much :)

badbrad2889:
it does it in order the way written already and yes I can change the "STD" to "standard" in the assets and the files if needed

badbrad2889:
@VladimirMakaev I reread and now think I understand what you are asking. You want it to be able to pick up all the gems at the same time instead of looping through?

badbrad2889:
@VladimirMakaev  im trying to resubmit the required changes. I completely understand what you are asking now so I will work on that for version 2. Thanks for the help and recomendations ;)

src/transmute/transmute.py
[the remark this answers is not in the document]
badbrad2889: this had been resolved. I see where this could cause issues and even crossed my mind when making originally. Thanks!

config/params.ini
badbrad2889: updated the ini to list style. Thanks

config/params.ini
```diff
@@ -111,6 +111,12 @@ gamble_items=
 stash_destination=3,2,1,0
 ;how often we want to run transmute routine(e.g. every 100 games)
 transmute_every_x_game=20
+; Add these possible gems to transmute them:
+; chipped
+; flawed
+; standard
+; flawless 
+transmute=chipped, flawed, standard, flawless
```
VladimirMakaev, line 119: Leave default as 'flawless' please. As it's not "free" to have a wider option

VladimirMakaev (changes requested):

badbrad2889:
I made the requested change Thanks.

VladimirMakaev (approved):
Can somebody test this before merging?

badbrad2889:
@mgleed @VladimirMakaev I made another change and optimize the gem transmute per recommendation. This is to decrease the "cost" and optimize the use of the inventory.

badbrad2889:
@mgleed Thank you for running the test again, 
@VladimirMakaev who is able to merge this request?

mgleed:
If one of the contributors can test this I'd be willing to merge. Not able to get on d2 til sunday to test myself

badbrad2889:
@mgleed ok sounds good, I appreciate it!

mgleed:
I didn't really know any good way to test this so i just threw together a character in hero editor with a bunch of amethysts. This is what happens. @VladimirMakaev @badbrad2889

https://user-images.githubusercontent.com/9866239/167376089-468028ba-cd25-4a4d-bcfb-b10762213794.mp4

mgleed (changes requested):
Refer to video comment in conversation.

mgleed:
I had tested this initially but realized the test case I'd contrived was wrong, leading to error. Deleted my video post from earlier. Tested this about three times with a variety of gems and it works nicely.

One way this could be improved is to include logic starting from chipped --> flawless for the number of gems to bring over to be cubed. I.e. if it knows we have 3 flawed gems and 2 normal gems, bring them all over and cube them into each other to make 1 flawless gem. But I don't think that needs to be included in this PR. Nice job @badbrad2889

badbrad2889:
@mgleed Thanks for retesting I greatly appreciate it! next implementation I will work on is going to be runes. after I will work on trying to have it grab all the gems at once if you are using that gem and the one above.

PR #742, opened 2022-05-02
Anya shopper now looks at all matches for a single template.

unixtreme:
General additional improvements.

### DIFF
--- a/src/shop/anya.py
+++ b/src/shop/anya.py
@@ -1,9 +1,9 @@
 import datetime
 import os
 import time
-import math
 
-import keyboard
+from collections import defaultdict
+
 import numpy as np
 
 from screen import grab, convert_screen_to_monitor
@@ -17,15 +17,7 @@ from utils.misc import wait, load_template
 from messages import Messenger
 
 
-def exit(run_obj):
-    run_time = str(datetime.timedelta(seconds=round(time.time() - run_obj.start_time)))
-    Logger.info("Exiting shopping mall...")
-    print(
-        "STATS \truns \t\ttime \t\tias_gloves_seen \tgloves_bought \tclaws_evaluated \tclaws_bought\n"
-        f"\t{run_obj.run_count} \t\t{run_time}"
-        f"\t\t{run_obj.ias_gloves_seen} \t\t\t{run_obj.gloves_bought} \t\t{run_obj.claws_evaluated} \t\t\t{run_obj.claws_bought}"
-    )
-    os._exit(0)
+ASSET_FOLDER = "assets/shop/gloves/"
 
 
 def wait_for_loading_screen(timeout):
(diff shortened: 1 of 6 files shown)


src/template_finder.py
```diff
@@ -63,9 +70,97 @@ def __new__(cls, save_last_res=False):
...
+        else:
+            template_match.center = ref_point
+            template_match.region = rec
+            template_match.valid = True
+        return template_match
+
     def get_template(self, key):
         return cv2.cvtColor(self._templates[key][0], cv2.COLOR_BGRA2BGR)
 
+    def search_v2(
```
unixtreme, line 97: Since this is my first contribution I didn't want to touch such an integral part of the program therefore I just made a `search_v2` method.
I'm not particularly excited by this approach but it will ensure the parts I touched don't interfere with the rest.
I think this logic is a bit easier to follow though.

src/utils/custom_mouse.py
```diff
@@ -228,6 +228,7 @@ def _move_to(x, y, absolute=True, duration=0):
         else:
             _winmouse.move_to(x, y)
 
+    @staticmethod
     def move(x, y, absolute: bool = True, randomize: Union[int, Tuple[int, int]] = 5, delay_factor: Tuple[float, float] = [0.9, 1.1]):
```
unixtreme, line 232: I don't know how this even worked before, but it should be broken.

src/template_finder.py
```diff
@@ -63,9 +70,98 @@ def __new__(cls, save_last_res=False):
...
+        image = image[ry:ry + rh, rx:rx + rw]
+
+        for idx, template in enumerate(templates):
+            if image.shape[0] > template.shape[0] and image.shape[1] > template.shape[1]:
+                res = cv2.matchTemplate(image, template, cv2.TM_CCOEFF_NORMED, mask=mask)
+                res = np.nan_to_num(res, copy=False, nan=0.0, posinf=0.0, neginf=0.0)
+                matches = np.where(res >= threshold)
+                matches = [pos for pos in zip(*matches)]
+                for coords in matches:
+                    # TODO: The TemplateMatch class should be able to construct itself with this information.
```
unixtreme, line 158: I'm still not super familiar with the codebase but I believe this assessment to be true.
I know it's a data class but if there's always logic associates with it maybe it should just be a normal object. Or not, who knows, lol.

src/shop/anya.py
```diff
@@ -106,61 +135,45 @@ def shop_loop(self):
...
                 for pos in claw_pos:
                     # cv2.circle(img, pos, 3, (0, 255, 0), 2)
                     x_m, y_m = convert_screen_to_monitor(pos)
-                    mouse.move(x_m, y_m, randomize=3, delay_factor=[0.5, 0.6])
+                    mouse.move(x_m, y_m, randomize=3, delay_factor=(0.5, 0.6))
                     wait(0.5, 0.6)
                     img_stats = grab()
                     trap_score = 0
                     melee_score = 0
+                    # TODO: Preload these templates.
```
unixtreme, line 176: I didn't go all the way with preloading all the templates because it seemed overkill, but looking at this and Drognan as well as some other request to add a third merchant made me think there would be a good argument to have a merchant base class with all the shared logic.

mgleed:
Thank you for your PR @unixtreme and sorry it's taken me some time to get back to you. I appreciate your work on this. There are a few open PR's regarding shopper that we're working on and I think this is the latest of three. My preference will be to merge #683 once stable. PR #676 has a lot of features like this one but a bit more comprehensive. Shop bot definitely needs work but all of us reviewers including myself haven't been able to catch up on everything yet. :p

PR #743, opened 2022-05-03
Make log chicken an option

drallcom3:
It can produce a lot of spam for low level bots.

### DIFF
--- a/config/params.ini
+++ b/config/params.ini
@@ -12,6 +12,7 @@ difficulty=hell
 info_screenshots=1
 loot_screenshots=0
 discord_status_count=20
+discord_log_chicken=1
 saved_games_folder=
 d2r_path=C:\Program Files (x86)\Diablo II Resurrected
 ; if you set this field to 1, botty will save the top-most visible character template just after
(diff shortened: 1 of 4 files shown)


D2RLegit:
I feel like this is more of a personal want then something we need congesting up in default params.

Aimware0:
Looks right to me

D2RLegit:
I would say no to this pr. Imo we don't need this in default botty. Chickens are seldom to none for the most part (outisde of the first 2 days of reset) where discord sends chicken logs shouldn't be a big deal to most.

bad-magick:
As a user I can say this would be very helpful. Since I run a few bots would be nice to be able to troubleshoot setups and have a better logs of what happens during gameplay. I experience chickens often. I’m sure there are other users that do too.

Bitty should be for both those with optimal and suboptimal setups. This would definitely be for those of us who don’t run perfect gear.

Just my thoughts. 😊

d

D2RLegit:
It already logs for chickens but default. Not sure how it would be helpful to have the ability to turn it off.

drallcom3:
I made it, because for bots with not so good items it produces a lot of messages. You might not actually care that your bots chickens 10% of the runs, but you don't lke that the chicken messages spam away the item found mesages.

This feature is not in any way super important though. Do with it whatver you see fit.

mgleed (approved):
Thank you for your work on this

PR #749, opened 2022-05-06
Walking Hammerdin bugfix

drallcom3:
The Hammerdin often accidentally clicked on monsters while walking to the end location. He did so while having Vigor active (because the bot thinks he's still walking). The existing fix to activate Concentration early did not work. Now it does.
You might have noticed many complaints about it in Discord.

If you want to test:
Do Eldritch or Trav 10x without the fix as a Hammerdin without any form of teleport. You will notice him casting hammers with Vigor active at the beginning.
With my fix that doesn't happen anymore.

Eld, Shenk, Trav, Pindle fixed. The others not, because they're impossible without teleport anyway.

### DIFF
--- a/src/char/hammerdin.py
+++ b/src/char/hammerdin.py
@@ -56,7 +56,7 @@ class Hammerdin(IChar):
         super().pre_move()
         # in case teleport hotkey is not set or teleport can not be used, use vigor if set
         should_cast_vigor = self._skill_hotkeys["vigor"] and not skills.is_right_skill_selected(["VIGOR"])
-        can_teleport = (self.capabilities.can_teleport_natively or self.capabilities.can_teleport_with_charges) and skills.is_right_skill_active()
+        can_teleport = self.capabilities.can_teleport_natively and skills.is_right_skill_active()
         if should_cast_vigor and not can_teleport:
             keyboard.send(self._skill_hotkeys["vigor"])
             wait(0.15, 0.25)
@@ -69,7 +69,7 @@ class Hammerdin(IChar):
 
     def kill_pindle(self) -> bool:
         wait(0.1, 0.15)
-        if self.capabilities.can_teleport_natively or self.capabilities.can_teleport_with_charges:
+        if self.capabilities.can_teleport_natively:
             self._pather.traverse_nodes_fixed("pindle_end", self)
         # If no teleport available -> Activate Concentration. Why? The bot will accidentally click on monsters and already cast hammers.
         else:
@@ -83,7 +83,7 @@ class Hammerdin(IChar):
         return True
 
     def kill_eldritch(self) -> bool:
-        if self.capabilities.can_teleport_natively or self.capabilities.can_teleport_with_charges:
+        if self.capabilities.can_teleport_natively:
             # Custom eld position for teleport that brings us closer to eld
             self._pather.traverse_nodes_fixed([(675, 30)], self)
         # If no teleport available -> Activate Concentration. Why? The bot will accidentally click on monsters and already cast hammers.
@@ -121,7 +121,7 @@ class Hammerdin(IChar):
         # Move a bit back and another round
         self._move_and_attack((40, 20), atk_len)
         # Here we have two different attack sequences depending if tele is available or not
-        if self.capabilities.can_teleport_natively or self.capabilities.can_teleport_with_charges:
+        if self.capabilities.can_teleport_natively:
             # Back to center stairs and more hammers
             self._pather.traverse_nodes([226], self, timeout=2.5, do_pre_move=False, force_tp=True, use_tp_charge=True)
             self._cast_hammers(atk_len)
--- a/src/char/hammerdin.py
+++ b/src/char/hammerdin.py
@@ -76,7 +76,7 @@ class Hammerdin(IChar):
             keyboard.send(self._skill_hotkeys["concentration"])
             wait(0.05, 0.15)
             # Traverse without pre_move, because we don't want to activate vigor when walking!
-            self._pather.traverse_nodes((Location.A5_PINDLE_SAFE_DIST, Location.A5_PINDLE_END), self, timeout=1.0, do_pre_move=False)
+            self._pather.traverse_nodes((Location.A5_PINDLE_SAFE_DIST, Location.A5_PINDLE_END), self, timeout=1.0, do_pre_move=False, force_tp=True, use_tp_charge=True)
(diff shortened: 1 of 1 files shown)


D2RLegit:
Not at a pc but it looks like you took out vigor altogether for the run and replaced it with concentration? Have you verified this still works with a hdin wearing a enigma?

drallcom3:
Vigor is still being use like before, up until the safe spot right before a boss. Just the last few meters are done with Concentration.
The bot was supposed to do that all along, but the functionality was broken.

Works fine with teleport charges.
I don't have an Enigma, but I don't see why it won't work. I haven't touched that part. Someone has to test it of course.

drallcom3:
I realized that I could just test it in singleplayer, so I did that. No issues with Enigma. Teleports happily around.

src/char/hammerdin.py
```diff
@@ -51,12 +51,6 @@ def pre_buff(self):
         mouse.click(button="right")
         wait(self._cast_duration, self._cast_duration + 0.06)
 
-    def on_capabilities_discovered(self, capabilities: CharacterCapabilities):
-        # In case we have a running pala, we want to switch to concentration when moving to the boss
-        # ass most likely we will click on some mobs and already cast hammers
-        if capabilities.can_teleport_natively:
-            self._do_pre_move = False
-
```
mgleed, line 59: Why was this function removed?

D2RLegit: I was wondering the same which is why I asked if it still works for hdin with tele. Ive been running as a walkadin and seems fine. Ill double verify in a bit if enigma still works.

D2RLegit:
[0.7.3-dev 2022-05-09 06:47:41,781] INFO       Capabilities: CharacterCapabilities(can_teleport_natively=False, can_teleport_with_charges=True)

Populates when using walkadin for pindle.
Walkadin does not work for pindle in this pr.

src/char/hammerdin.py
drallcom3: When debugging self._do_pre_move it turned out that it reverts back to true. The Hammerdin class logic expected it to stay false though. It was the main culprit of why the Hammerdin seemed to "attack" while having Vigor turned on. It should have switched to Concentration, but didn't. Not relying on self._do_pre_move fixed it immediately and then the on_capabilities_discovered wasn't required anymore.

Also it lead to unneccesarily complicated logic constructs that were hard to follow and could be done in a simpler and more readable way.

drallcom3:
I get:
INFO       Capabilities: CharacterCapabilities(can_teleport_natively=False, can_teleport_with_charges=False)
and when adding checks further down the line:
DEBUG      capabilities.can_teleport_natively: False
DEBUG      capabilities.can_teleport_with_charges: False
when doing Pindle without any teleport equipped.

Did you get that message while having a teleport charge item equipped and ready?

I can only achieve
INFO       Capabilities: CharacterCapabilities(can_teleport_natively=False, can_teleport_with_charges=True)
if I have a teleport item with charges equipped and a hotkey assigned in-game and a hotkey assigned in the config. Which is (I assume) cirrect behaviour. If you did the test with a teleport charge item equipped, you're not a Walkadin anymore. I just ran the test. With teleport charges the Hammerdin behaves exactly like in master. Which is correct behaviour, since this PR is about walking, not teleport charges. The Hammerdin needs some more fixes, as you have discovered, but I didn't want to make this PR too massive.

In short: I can't see a bug here (that didn't already exist).

D2RLegit:
It saying i have teleport charges when i run pindle even though im not using any teleport or teleport charge items. it should be a walkdin and its being recognized that it has teleport charges.

drallcom3:
Thanks.
If your self.capabilities.can_teleport_with_charges were incorrectly true, that would be a bug well beyond the Hammerdin file (as it's not being set in there and the info shows up before any boss functions are executed).

D2RLegit:
Pretty sure its because you removed 

    def on_capabilities_discovered(self, capabilities: CharacterCapabilities):
        # In case we have a running pala, we want to switch to concentration when moving to the boss
        # ass most likely we will click on some mobs and already cast hammers
        if capabilities.can_teleport_natively:
            self._do_pre_move = False

but i could be wrong. My walkdin works fine on every other pr.

drallcom3:
What exactly does happen to your Hammerdin when you do Pindle with this PR?

drallcom3:
It's actually this #758 bug.
The Hammerdin just made us discover it.

src/char/hammerdin.py
VladimirMakaev: I wouldn't touch this code. It can easily break something for Enigma / Tele-Charge paladin . And you need to test 3 different version of paladin on all routes to make sure it's sound:
1) Walkdadin Trav/Shenk/...
2) Enigma Paladin Trav/Shenk/...
3) Tele-Charge Paladin Trav/Shenk

VladimirMakaev: You should just cast Concentration and move with PreMove=False that way you won't change to Vigor. 
preMove() will flip to Vigor in case of Walkadin or Tele-Charge (there is no way to pass that telecharge is to be used based on current code structure). So in case you're opting to tele-charge then you should manage PreMove part manually. E.g. casting concentration and doing PreMove=False, TeleCharge = True or PreMove=True if you want Vigor to be on for walking / running out of tele charges and walking.

VladimirMakaev (changes requested):

src/char/hammerdin.py
drallcom3: > I wouldn't touch this code. It can easily break something for Enigma / Tele-Charge paladin . And you need to test 3 different version of paladin on all routes to make sure it's sound:
> 
>     1. Walkdadin Trav/Shenk/...
> 
>     2. Enigma Paladin Trav/Shenk/...
> 
>     3. Tele-Charge Paladin Trav/Shenk

It didn't even work the way it was intended to work. No need to re-introduce bugs.

This PR has been tested extensively with all three variants. Some people even happily and successfully use it for their daily runs.

drallcom3: > You should just cast Concentration and move with PreMove=False that way you won't change to Vigor. preMove() will flip to Vigor in case of Walkadin or Tele-Charge (there is no way to pass that telecharge is to be used based on current code structure). So in case you're opting to tele-charge then you should manage PreMove part manually. E.g. casting concentration and doing PreMove=False, TeleCharge = True or PreMove=True if you want Vigor to be on for walking / running out of tele charges and walking.

Which is what I did.

PR #763, opened 2022-05-12
Initial Fohdin

jobithu:
Only does Pindle, start of Trav
Made base Paladin
FoHdin inherits from Paladin

now adding chaos sanctuary


mgleed:
Closed due to #765

PR #770, opened 2022-05-14
Refactor: TemplateFinder()

mgleed:
Main changes are in template_finder.py, the remaining changes are to modify the codebase accordingly.

Use a Template dataclass object to store template information rather than a list (before it was like self._templates[key][0] to get image, [key][1] to get grayscale, etc.)

Deconstruct TemplateFinder class

Use @cache to store all templates in _templates().

Extracted a lot of the logic from template_finder.search() into template_finder._single_template_match(), which is now called from search().

Added new function search_all(), which will need integration into shopper bots as separate PR. This new function is included in template_finder_test.py and test passes.

Modify dataclass for TemplateMatch to include normalized monitor coordinates as attributes so convert_screen_to_monitor and normalize_monitor stuff doesn't need to be used.


mgleed:
Requesting seasoned devs to review. @aeon0 or @VladimirMakaev if you get a chance to review the changes to template_finder.py and give me feedback that'd be appreciated. Thanks!

src/template_finder.py
```diff
@@ -9,235 +9,222 @@
...
+                img_gray = cv2.cvtColor(i, cv2.COLOR_BGR2GRAY),
+                alpha_mask = alpha_to_mask(i)                    
+            ))
+    return templates
+
+def _grayscale_template_and_image(template: Template, img: np.ndarray) -> tuple[np.ndarray]:
+    return template.img_gray, cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
+
+def _color_filter_template_and_image(template: Template, img: np.ndarray, color_match: list = None) -> tuple[np.ndarray]:
+    return color_filter(template.img_bgr, color_match)[1], color_filter(img, color_match)[1]
```
aeon0, line 89: I really like the functional approach taken. But for these two functions that are used once in a pretty obvious if statement, I dont see the benefit.

aeon0 (changes requested):
Really like it 😍 I wanted to do pretty much exactly that refactoring to be able to use multiprocessing more easily on the template searching.

mgleed:
> Really like it 😍 I wanted to do pretty much exactly that refactoring to be able to use multiprocessing more easily on the template searching.

Thanks! Any concerns about thread safety?

aeon0:
The only thing I am a bit worried about is the cached `_template()` which is not threadsafe. Maybe put a lock around it in `get_template()` and always call that one.

aeon0 (approved):
Noice!

mgleed:
Closes #683

PR #775, opened 2022-05-16
Reduce identify time by ~8 seconds

reqyl:
Previously the cain dialog would stay open blocking inventory key from working (when checking items), the bot would be in some loop for 8 seconds before closing it and continuing.


derek337:
This works well, I tested locally. Much faster.

mgleed:
This used to be on master but there are a few cases where it fails. I don't remember what they were now. Tried the return to play method but even that doesn't work for everyone so had to increase the time. Best is to use an ID tome til someone makes a permanent solution.

derek337:
> This used to be on master but there are a few cases where it fails. I don't remember what they were now. Tried the return to play method but even that doesn't work for everyone so had to increase the time. Best is to use an ID tome til someone makes a permanent solution.

How does ID tomes work? Will the program just attempt to use it if present in character inventory?

mgleed:
Yes

reqyl:
I checked the git history for this file and it look to me that the code was last changed here #582

The previous behavior seems to have been that the dialog wasn't closed at all (maybe after some timeout somewhere though)

So I think this change is still relevant and correct. In the current code some loop inside `view.return_to_play()` is what is taking the extra 8 seconds.

mgleed:
> Previously the cain dialog would stay open blocking inventory key from working (when checking items),

Hey I'm trying to see if I can fix the original return_to_play() function. Could you capture this on video or with screenshots?

mgleed:
I couldn't reproduce the issue, but I can tell where it was likely going wrong. The NPCMenu often disappears for a fraction of a second before reappearing after hitting the identify button which likely caused issues with the view.return_to_play() function. I added a couple small adjustments but this should work now.

mgleed (approved):

mgleed (changes requested):
looks like your rebase / force push reverted some changes from master, gonna see if i can salvage the changes

mgleed:
Had to open new PR #837

PR #777, opened 2022-05-17
Walking poisonmancer hc setup

D2RLegit:


D2RLegit:
DO NOT MERGE to master.
PR is for walkamancer only. It will disrupt all other builds for eld/shenk/trav if merged. If someone would like to convert this to work with master that would be great, this will also fix walkadin when walking on stairs to get to shenk with minions in the way. However, I dot not have time to convert it for all chars. nova necro gets stuck often during poison nova and requires to move via traversing to nodes only in order to work. I may have found a solution to fix walking sticking in poison_nova.py however the best results come from layering traversing from the run.py's. So the script looks super funky but is very functional.

D2RLegit:
Closing due to not needing to be an open pr

PR #827, opened 2022-06-03
Update transmute.py

xw220:
If max game length reached when you're transmuting, the game will be forced quitted. There may be some perfect gems or flawless gems left in the cube that cause all the later transmuting will be skipped. To resovle this problem, I add clear_cube() function to put all gems in cube back to stash before gem picking.


mgleed:
seems reasonable to me. @VladimirMakaev will you take a peek?

VladimirMakaev:
Several issues here:
1) If you have less slots in your inventory than in the cube this will fail. Probably need more robust way of transferring items from cube to the stash
2) I don't like that it's checking this prior every time that means happy path which is a majority of the time will be slower. I'd rather have a logic that detects that cube isn't empty during a transmute that would force to put everything back in the stash and then clear the cube.

But it doesn't have to be perfect. Feel free to merge if you're ok with mentioned concerns

xw220:
Bug fixed. After pick gems from cube, I forgot to put gems back to stash which will cause gempicking error. @_@

mgleed:
Thanks for this @xw220 . Do you think you could implement the second suggestion from Vladimir's review? His first concern would require revision of a lot of other code, so no worries there. Thank you @VladimirMakaev for the review by the way

mgleed (changes requested):
see above

xw220:
I'll try.

xw220:
I noticed transmute.py is updated. Close this for now, and I'll make change on new tansmute.py.

PR #829, opened 2022-06-05
Optimize/trav

reqyl:
- Pass list of run scripts in chronological order to each type of run. This allows for optimizations. This kind of optimization is added to run_trav when run_trav is the last script. Skip teleporting into position so that tp is not blocked, simply just exit instead.

- Trav optimization: Prefer to not use tele charges when getting back inside to loot when using charges. Only use tele charge if pathway was blocked. This change makes the bot use less tele charges, and thereby improve run speed due to not having to repair as often. If you can tele natively the change will not affect behavior.

### DIFF
--- a/src/bot.py
+++ b/src/bot.py
@@ -101,17 +101,17 @@ class Bot:
         # Adapt order to the config
         self._do_runs = OrderedDict((k, self._do_runs[k]) for k in Config().routes_order if k in self._do_runs and self._do_runs[k])
     
-        last_run = list(self._do_runs.keys())[-1]
+        runs = list(self._do_runs.keys())
         self._do_runs_reset = copy(self._do_runs)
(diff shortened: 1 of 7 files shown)


mgleed (changes requested):
I think this is a good idea, thank you.

Could I suggest that rather than passing `last_run` we just pass `runs`? Then just change the logic for `_is_last_run` in trav.py.

That way if down the line we ever want to check things like `_is_first_run` or check on what run is next we could do it on the fly.

mgleed (approved):
Looks good. Can you confirm this works in game after changes? I'm not in a place to test myself atm.

reqyl:
Yep did ~50 trav runs just now with the latest change applied works fine.

I've had these optimizations (and a few more) for a long time but haven't gotten around do make a PR

PR #832, opened 2022-06-06
Move _lerp function to i_char.py

roozhou:
There is no reason to have 3 copies of the same function. Let's make _lerp a common function for all characters.


mgleed (approved):
Thank you

PR #833, opened 2022-06-06
Bugfix: Random guess should be random

roozhou:
"Random guess" is always towards bottom when blocked.


mgleed (approved):
Good catch, thank you

Issue #789, opened 2022-05-20
Suggestion: mouse click over online tab before finding/clicking play button


Tntrance:
**Is it really a Bug report?**
If you have questions regarding setup or feature information/requests, please consider asking in the Discord channel: https://discord.gg/Jf3J8cuXWg. The Github issues section is for bug reports only.

**Botty Version**
What version of botty are you running? Are you using a pre-compiled release? Which one. Are you using the latest master branch? On which commit?

**Describe the bug**
I’ve noticed that as of recent the bot keeps getting stopped at the play button due to blizzards lack of server upkeep. After a game has ended the tab on the right will default to Offline characters instead of Online

**To Reproduce**
1. Its random. Whenever Bnet or your own computer is having network issues.
2. Script will stop at looking for play button phase

**Expected behavior**
Bot will mouse hover and click on “Online” tab on the right hand side. Then continue to look for play button.

**Screenshots / Video**
If applicable, add screenshots or even better a short video of the error happening to help explain your problem. Botty often automatically makes screenshots when something goes wrong. Add these also.

**Logs**
Add the relevant part of the info.log in this section. Either upload it as a fail or copy-paste the relevant part in here.

**Additional info**
Add any other information about the problem here.

zutto:
`This project is for informational and educational purposes only and is not meant for online usage.`

first textline in https://github.com/aeon0/botty/blob/master/README.md

roozhou:
https://github.com/aeon0/botty/pull/834

PR #834, opened 2022-06-06
Fix Issue #789: Select online tab before creating game when d2r switch to offline tab automatically

roozhou:
Added option `online_char` to params.ini.
Will check online tab status before creating games when `online_char` is 1.

### DIFF
--- a/config/params.ini
+++ b/config/params.ini
@@ -19,7 +19,7 @@ d2r_path=C:\Program Files (x86)\Diablo II Resurrected
 ; if you set this field to 1, botty will save the top-most visible character template just after
 ; the online/offline tab and upon restart will attempt to find the same template to select the char
 restart_d2r_when_stuck=0
-
+online_char=0
 
 [routes]
 ; Add these possible routes to "order" to run them:
--- a/config/params.ini
+++ b/config/params.ini
@@ -5,7 +5,7 @@ message_api_type=discord
 custom_message_hook=
 custom_loot_message_hook=
 max_game_length_s=380
-max_consecutive_fails=0
+max_consecutive_fails=5
 max_runtime_before_break_m=0
 break_length_m=0
 randomize_runs=0
@@ -16,10 +16,8 @@ discord_status_count=20
 discord_log_chicken=1
 saved_games_folder=
 d2r_path=C:\Program Files (x86)\Diablo II Resurrected
-; if you set this field to 1, botty will save the top-most visible character template just after
-; the online/offline tab and upon restart will attempt to find the same template to select the char
+; if you set this field to 1, botty will attempt to restart d2 after a crash or failure
 restart_d2r_when_stuck=0
-online_char=0
 
 [routes]
 ; Add these possible routes to "order" to run them:
--- a/src/bot.py
+++ b/src/bot.py
@@ -30,7 +30,7 @@ from char.poison_necro import Poison_Necro
 from char.bone_necro import Bone_Necro
 from char.basic import Basic
 from char.basic_ranged import Basic_Ranged
-from ui_manager import wait_until_hidden, wait_until_visible, ScreenObjects, is_visible
+from ui_manager import wait_until_hidden, wait_until_visible, ScreenObjects, is_visible, detect_screen_object
 from ui import meters, skills, view, character_select, main_menu
 from inventory import personal, vendor, belt, common, consumables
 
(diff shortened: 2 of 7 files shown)


mgleed:
> Added option `online_char` to params.ini.
> Will check online tab status before creating games when `online_char` is 1.

This is semi implemented with game_recovery.py

On botty start it saves online status and saves template of the character so if game crashes it'll reselect the correct character. Sounds like this function could be extended for this purpose.

roozhou:
> > Added option `online_char` to params.ini.
> > Will check online tab status before creating games when `online_char` is 1.
> 
> This is semi implemented with game_recovery.py
> 
> On botty start it saves online status and saves template of the character so if game crashes it'll reselect the correct character. Sounds like this function could be extended for this purpose.

This change reuses `select_online_tab` in character_select.py, which is also used by game_recovery.py. Something different on this issue:
1) Game does not crash.
2) It may switch back to offline even after you click the online tab. The first try always fails when it happens to me. Checking the active tab is always needed before clicking the play button.

mgleed:
> > > Added option `online_char` to params.ini.
> > > Will check online tab status before creating games when `online_char` is 1.
> > 
> > This is semi implemented with game_recovery.py
> > 
> > On botty start it saves online status and saves template of the character so if game crashes it'll reselect the correct character. Sounds like this function could be extended for this purpose.
> 
> This change reuses `select_online_tab` in character_select.py, which is also used by game_recovery.py. Something different on this issue:
> 1) Game does not crash.
> 2) It may switch back to offline even after you click the online tab. The first try always fails when it happens to me. Checking the active tab is always needed before clicking the play button.

Agree this logic needs fixing. But if we detect the online / offline status on startup why do we need a param for it?

roozhou:
> game_recovery.py

I don't think game_recovery.py checks online/offline status. Currently checking online status is performed in [on_select_character](https://github.com/aeon0/botty/blob/master/src/bot.py#L224) in bot.py only when `restart_d2r_when_stuck` is enabled.

Botty can also be started inside the game. What if BN disconnects after the first game?

Using `restart_d2r_when_stuck` to control online/offline status is also confusing.

mgleed:
> > game_recovery.py
> 
> I don't think game_recovery.py checks online/offline status. Currently checking online status is performed in [on_select_character](https://github.com/aeon0/botty/blob/master/src/bot.py#L224) in bot.py only when `restart_d2r_when_stuck` is enabled.
> 
> Botty can also be started inside the game. What if BN disconnects after the first game?
> 
> Using `restart_d2r_when_stuck` to control online/offline status is also confusing.

Agree it's confusing. And you're right, it's not checked there, it's just checked on startup. 

Would this solve everything?
```
    def on_select_character(self):
-        if Config().general['restart_d2r_when_stuck']:
            # Make sure the correct char is selected
            if character_select.has_char_template_saved():
                character_select.select_char()
            else:
                character_select.save_char_online_status()
                character_select.save_char_template()
        self.trigger_or_stop("create_game")
```

roozhou:
> Would this 1-liner solve everything?
No. This assumes your online/offline tab status is correct on your first game creation. This is not guaranteed since you are allowed to start botty inside the game.

Imagine you start botty in a game. The server lags and brings you to offline chars after exit. Botty sticks to offline thereafter.

mgleed (approved):
Thank you for this PR. I tested it with crashes in a few ways and found errors with game recovery that I tried to fix in this PR. I took out the online param because the only reason it would ever be necessary is if the game crashed on the very first run AND botty was launched in town. Botty appropriately detects online status for you in every other scenario.

mgleed:
Fixes #789

DongGamHe:
[0.7.3-dev 2022-06-08 15:55:35,564] INFO       Character select failed. Shut down botty
Exception in thread Thread-7 (start):
Traceback (most recent call last):
  File "threading.py", line 1009, in _bootstrap_inner
  File "threading.py", line 946, in run
  File "bot.py", line 163, in start
  File "bot.py", line 183, in trigger_or_stop
  File "transitions\core.py", line 899, in _get_trigger
  File "transitions\core.py", line 401, in trigger
  File "transitions\core.py", line 1201, in _process
  File "transitions\core.py", line 426, in _trigger
  File "transitions\core.py", line 435, in _process
  File "transitions\core.py", line 272, in execute
  File "transitions\core.py", line 1123, in callbacks
  File "transitions\core.py", line 1144, in callback
  File "bot.py", line 226, in on_init
  File "bot.py", line 192, in restart_or_exit
  File "utils\restart.py", line 21, in safe_exit
  File "utils\restart.py", line 25, in kill_game
  File "utils\restart.py", line 14, in process_exists
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc0 in position 2: invalid start byte


I have a utf8 character and I get this error when I run the botty after the update today.

PR #836, opened 2022-06-07
Fix generic_api

xw220:
In generic_api.py, custom_loot_message_hook was not supported. I fixed it and it works fine now. If custom_loot_message_hook isn't set in params.ini, use custom_message_hook instead.

### DIFF
--- a/src/messages/generic_api.py
+++ b/src/messages/generic_api.py
@@ -8,7 +8,7 @@ class GenericApi:
 
     def send_item(self, item: str, image:  np.ndarray, location: str, ocr_text: str = None):
         msg = f"Found {item} at {location}"
-        self._send(msg)
+        self._send(msg, loot = True)
 
     def send_death(self, location: str, image_path: str = None):
         msg = f"You have died at {location}"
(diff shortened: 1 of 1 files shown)


src/messages/generic_api.py
```diff
@@ -32,14 +32,10 @@ def send_message(self, msg: str):
     def _send(self, msg: str, loot = False):
         msg = f"{Config().general['name']}: {msg}"
 
-        if loot:
```
mgleed: I simplified the logic here. Does the difference make sense to you? The walrus operator `:=` is great for this sort of scenario--if you haven't learned about it yet check it out.

src/messages/generic_api.py
```diff
@@ -32,14 +32,10 @@ def send_message(self, msg: str):
...
-            url = Config().general['custom_loot_message_hook']
-            if not url:
-                url = Config().general['custom_message_hook']
-        else:
-            url = Config().general['custom_message_hook']
-        if not url:
+        if not (url := Config().general['custom_message_hook']):
             return
+        if loot and (loot_url := Config().general['custom_loot_message_hook']):
+            url = loot_url
```
mgleed, line 38: I simplified the logic here. Does the difference make sense to you? The walrus operator := is great for this sort of scenario--if you haven't learned about it yet check it out.

xw220: Much better than mine. :)

mgleed (approved):

PR #847, opened 2022-06-09
Adding the rest of sazabis to pickit options.

NeoIsrafil:
All images are ready and working, pickit file is set with 0 as default but activatable. Figured it'd be nice to have the rest since its good for barb mercs.


NeoIsrafil:
I'm dumb and didn't include the PNGs along with the jpegs.... meant for you guys to have BOTH just incase youd need em.  Gonna see if it lets me edit

NeoIsrafil:
I'll re-add this once i get a chance to fix it... Something is VERY wrong, it has decided that EVERY item is saabis helmet now, and i have no clue why because it was working earlier....

PR #849, opened 2022-06-09
Added support for trapsin to tele trav

badbrad2889:
https://streamable.com/zhjiwz


badbrad2889:
> Why is native teleport required since very limited number of tp is used? TP charges should be good as well.

you are right I can make it for all tele including charges, up to you

badbrad2889:
I added support for the tele with charges.

Riddle1001:
Add variation to the wait @badbrad2889

PR #853, opened 2022-06-10
Pickit click backtrack

jobithu:
When looting multiple items in high monster density areas (e.g. Chaos Sanctuary), Botty sometimes gets "off-track" and ends up at a place, where no node is visible. In these cases, the run will fail, as pather gets stuck.

To avoid the failed games, the idea is to record the x,y coordinates for each click made during a pickit routine in a stack (click_history) and pop the stack after the pickit routine is finished - thereby bringing us (not directly, but roughly) back to the original position where pickit started.

Ideally this avoids failed runs.


jobithu:
To be merged to NIP, not Master ;)

jobithu:
take care, this PR uses a modified default pickit (sosa,gleed,legit V5 dated 11JUN)

jobithu:
This PR will be no longer needed when automap-based pathing is implemented.
Closing the PR for now to focus on automap instead.

PR #857, opened 2022-06-11
Added new param: Prefer_act4_town

jobithu:
The most efficient way to farm Trivial is to end the run in A4, instead of A3.
The subsequent runs will be faster, as A3 town is rather large, and traverse to WP & Vendors takes longer.

Setting "prefer_act4_town" to 1 will trigger a TP at the end of Travincal, and our char traversing to A4 town to end the run there.

Intensive testing on a fully geared Enigma Hammerdin revealed that average run duration decreased from 82 to 71 seconds (12% improvement). For a typical farming day of 12h, instead of completing 525 runs, additional 85 runs can be performed (total 610 runs).

The methodology can be adapted to other runs such as Arcane, the benefit here, will however be much less.

### DIFF
--- a/config/params.ini
+++ b/config/params.ini
@@ -62,6 +62,7 @@ battle_command=f8
 stash_gold=1
 min_gold_to_pick=1000
 use_merc=1
+prefer_act4_town=1
 ; Attack length for barbarians should be as high as 8-10 and even 10-12 for trav/shenk
 atk_len_arc=2.5
 atk_len_trav=3.0
--- a/src/config.py
+++ b/src/config.py
@@ -307,6 +307,7 @@ class Config:
             "gamble_items": False if not self._select_val("char", "gamble_items") else self._select_val("char", "gamble_items").replace(" ","").split(","),
             "sell_junk": bool(int(self._select_val("char", "sell_junk"))),
             "enable_no_pickup": bool(int(self._select_val("char", "enable_no_pickup"))),
+            "prefer_act4_town": bool(int(self._select_val("char", "enable_no_pickup"))),
         }
         # Sorc base config
         sorc_base_cfg = dict(self.configs["config"]["parser"]["sorceress"])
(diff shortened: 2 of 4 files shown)


namdnas23:
I run pindle eld shenk trav then nihlathak for this reason. Ends back in act 5 so it traverses less distance

D2RLegit:
2 extra load screens for this even though they may somewhat quick sometimes. Not sure if its really cuts that much time for most people due to the extra loading screens/sell items/stashing/pickit and especially on vm's. Also when joining a new game you typically walk to vendor to resupply or sell items which ormus is roughly similar distance as jamal when starting a5. Just my 2 cents.  It better to start in act 5 and end in act 5 aka start with pindle and end with eld/shenk or nihlathak.

jobithu:
Thanks for your feedback, Legit!

Due the points you mentioned, the default value of the param is 0 - in case multiple targets are farmed, I agree with you: we should end up in A5 instead.

Based on your feedback, I did pressure tests on VM and as Walkadin with / without the param activated, just farming Trav. 
Even taking into account the extra loading screens, my average run duration was 21s shorter 85 more games in 12h):
param off = Avg Game Length: 00:01:58 (120 games)
param on = Avg Game Length: 00:01:37 (200 games)

Looking at it: I would definitely merge it :)

mgleed:
I'd like to incorporate a feature like this, but one that is more generalizable. Like a way to end in whatever act you put in a param.

mgleed (changes requested):
see prior msg

jobithu:
So for Arcane, it could also make sense to just pop the tome, when at summoner and leave through the WP of Duriel. This could allow you to end up in A4 town, as well (when powerfarming hate keys). 

so in this case: the param would be: preferred_end_town= "A1, A2, A3, A4, A5". Relevant runs: Trav & Arcane.

mgleed:
This is not ready to merge. It says "end_run_in_act" which implies you want botty to force to end the run in a certain act regardless of what run you're on, but this is not at all implemented. The only thing it does is if trav is your last run it'll end the run in X act.

PR #865, opened 2022-06-13
fix: handle keyerror during backup settings

erikmangsten:
Keyerror when `["AdditionalLaunchArguments"`] is missing during backup settings. Proposing that it should just pass since you can't do anything about it at that moment.

### DIFF
--- a/src/utils/auto_settings.py
+++ b/src/utils/auto_settings.py
@@ -44,10 +44,13 @@ def backup_settings():
             return
     f = open(current_file)
     curr_settings = json.load(f)
-    launch_options = curr_settings["Games"]["osi"]["AdditionalLaunchArguments"]
-    with open(backup_file, 'w') as f:
-        f.write(launch_options)
-    print("D2R launch options backed up successfully.")
+    try:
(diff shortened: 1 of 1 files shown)


Riddle1001:
If you can put a Logger.warning saying that the botty failed to set the launch options and possibly tell the user how to solve the issue

src/utils/auto_settings.py
```diff
@@ -44,10 +44,13 @@ def backup_settings():
...
-    with open(backup_file, 'w') as f:
-        f.write(launch_options)
-    print("D2R launch options backed up successfully.")
+    try:
+        launch_options = curr_settings["Games"]["osi"]["AdditionalLaunchArguments"]
+        with open(backup_file, 'w') as f:
+            f.write(launch_options)
+        print("D2R launch options backed up successfully.")
+    except KeyError:
+        pass
```
Riddle1001, line 53: Put a Logger.warning saying that botty failed to set the launch options and possibly tell the user how to solve the issue, other than that looks good :D

Riddle1001 (changes requested):

PR #867, opened 2022-06-13
WIP: Automap pather. Fix walking to nodes instead of teleporting

roozhou:
Use automap to locate the char.

To add automap support for other runs / towns:
1. Find unique objects on automap. Unreachable objects preferred since char/merc crosses on automap interfere with template matching.
2. Cut objects from screenshots, delete non-map regions and make those pixels transparent.
3. Start pather.py, run to the nodes and press F11 to find the offset to each objects.
4. Add nodes to self._nodes in pather.py


jobithu:
So as a proof of concept: it works.

Next steps would be to 
- add more nodes and templates to existing nodes to make pathing a bit more consistent & independent of map layout. 
- add the remaining nodes for bosses in fohdin.py (kill_<sealboss>_automap())
- speed up  the minimap based pathing (currently quite slow 1 teleport / second) -> cuda core multithreading (Nvidia only)?
- implement the cs_rush, cs_kill_trash pathing & logic (currently in diablo_old.py, but I was too lazy to complete it yet).
- implement additional nodes for the cs_kill_trash and walkadin versions.

jobithu:
Dia ready for testing
FOHdin only.
Requires teleport.

All towns implemented, but A1 town having some issues with layout check of waypoint: requires better templates.

i would love to have your data in pressure tests for Diablo. If you would like to support, please 1) delete the content of your log folder 2) run only Diablo (not other bosses), starting from A4 town. 3) maxgamelength of 360s should be fine 4) after having finished your runs, please use utils/diablo_log_parser.py and upload here the file the script created (stored in your botty root folder).

Any other feedback and observations are welcome!

aliig (commented):
Just one thing I've noticed. You should strive to have as little alpha as possible on your assets. You'll end up with lots of false positives if you have a lot of transparency in your asset.

Template:
A1_TOWN_AUTOMAP_NORTH

Match:
image

jobithu:
> Just one thing I've noticed. You should strive to have as little alpha as possible on your assets. You'll end up with lots of false positives if you have a lot of transparency in your asset.
> 
> Template: A1_TOWN_AUTOMAP_NORTH
> 
> Match: image

indeed this is not the strongest template, A1 town is broken for the north WP. But the matches you see will have a low probability - the alphamapped pictures will have to match with >0.95 threshold.

mgleed:
> > Just one thing I've noticed. You should strive to have as little alpha as possible on your assets. You'll end up with lots of false positives if you have a lot of transparency in your asset.
> > Template: A1_TOWN_AUTOMAP_NORTH
> > Match: image
> 
> indeed this is not the strongest template, A1 town is broken for the north WP. But the matches you see will have a low probability - the alphamapped pictures will have to match with >0.95 threshold.

Just saying having a lot of alpha in your assets is going to cause problems. I picked the first example of many. We had to revert template matching for an empty merc frame (alpha'd except for border) because people's platforms were interpreting the asset differently. @aeon0 probably remembers. ;)

aeon0:
Haha, I do remember :D

D2RLegit:
Noticed that starting in act 1 and act 3 get stuck and fail game. Haven't tried act 2 yet act 4 and 5 seem to work.

jobithu:
> Noticed that starting in act 1 and act 3 get stuck and fail game. Haven't tried act 2 yet act 4 and 5 seem to work.

Can you add info screenshots and logs so I can investigate and provide a fix ?

moor3z:
[log.txt](https://github.com/bottytools/botty/files/8995328/log.txt)
Log File - 12 Hours

D2RLegit:
Town visit still broken. Not sure how to fix.

jobithu:
> Town visit still broken. Not sure how to fix.

reason is that the automap check does not work properly (see here: https://github.com/bottytools/botty/blob/8856ef8cc8137e90ad6c3ec1324ef3d90905834e/src/automap_finder.py#L48). so it does not correctly detect if the minimap is on. so all the functions in a4 town.py need a toggle for automap before and after. whilst this works well for routine play, it begins to cause issues once we use cs_town_visits.

moor3z:
[log.txt](https://github.com/bottytools/botty/files/9005921/log.txt)
- 
Log - FoH Paladin 
**8 Hours -- 15 Failures**
Pull Request #867

PR #905, opened 2022-06-18
Feature/auto detect hotkeys

Ezro:


src/utils/hotkeys.py
```diff
@@ -0,0 +1,354 @@
...
+from config import Config
+from ui.skills import get_selected_skill
+from screen import convert_screen_to_monitor, grab
+import template_finder
+import keyboard
+from utils.custom_mouse import mouse
+from utils.misc import wait
+from logger import Logger
+
+d2r_keymap = {}
```
aeon0, line 12: Please add a short comment on what these are. Only needed for the "public" ones. Thus if anybody wants to use this data, they can look it up here.

src/utils/hotkeys.py
```diff
@@ -0,0 +1,354 @@
...
+        templates,
+        key_list,
+        starting_left_skill,
+        starting_right_skill,
+        found_keys,
+        _default_left_skill_key_map,
+        _default_right_skill_key_map)
+    for key in found_keys:
+        if key in key_list:
+            key_list.remove(key)
```
aeon0, line 120: Why we need to do this twice? Also if we have to, use a for loop.

src/utils/hotkeys.py
```diff
@@ -0,0 +1,354 @@
...
+    if Config().char['cta_available'] and HotkeyName.SwapWeapons in d2r_keymap:
+        keyboard.press_and_release(d2r_keymap[HotkeyName.SwapWeapons])
+        wait(0.3)
+        img = grab()
+        starting_left_skill = get_selected_skill(templates, img, Config().ui_roi["skill_left"])
+        starting_right_skill = get_selected_skill(templates, img, Config().ui_roi["skill_right"])
+        _find_keymapping(
```
aeon0, line 127: SAme as above.

src/ui/skills.py
```diff
@@ -10,6 +11,196 @@
 from ui_manager import wait_until_visible, ScreenObjects
 from d2r_image import ocr
 
+class SkillName(str, Enum):
```
aeon0, line 14: These names have to map the ones used in the skill templates, right? Maybe a comment would be nice about this kind of hidden dependency.

src/template_finder.py
```diff
@@ -222,6 +249,27 @@ def search_all(
             break
     return matches
 
+def search_all_templates(
```
aeon0, line 252: Why can you not use search_all() here?

src/template_finder.py
```diff
@@ -60,10 +60,37 @@ def stored_templates() -> dict[Template]:
             )
     return templates
 
+@cache
+def stored_templates_by_dir() -> dict[Template]:
```
aeon0, line 64: I dont like that we bascially copy the stored_templates() functionality here. I'd rather go for adding the dir normally to TEMPLATE_PATHS and to get all the specific template names loop through folder and get names of the files in a list.

aeon0: Then we also can ditch the search_all_templates() function.

src/template_finder.py
aeon0: If we really need the list[Template] as input, we should add it to the serach_all() function.

aeon0 (dismissed):

mgleed:
One thing that needs to be tested here is localization. Botty currently allows users to select characters that have names that aren't in English alphabet. Not sure if this will break things if people have a keyfile with kanji, for example.

src/utils/hotkeys.py
Ezro: Comments have been added for describing the variables

src/utils/hotkeys.py
Ezro: We do this twice in order to try to confidently deduce that the key that was pressed has actually changed the left / right skill slot. This has been abstracted into a _deduce_hotkeys helper fn

src/ui/skills.py
Ezro: Class docstring has been added to SkillName and HotkeyName

src/template_finder.py
mgleed: This suggests shape of img and roi are incompatible. 

But rather than going down that rabbit hole we should just refactor search_all(). Alex pointed out there's a better way with cv2. https://discord.com/channels/904057828791185458/906298605420285993/978388103800713316

Ezro:
> One thing that needs to be tested here is localization. Botty currently allows users to select characters that have names that aren't in English alphabet. Not sure if this will break things if people have a keyfile with kanji, for example.

image
I've confirmed that this should work for any provided language, but the Logger doesn't seem to like the encoding; I can update the debug line to "Determine hotkeys" and remove the key_name to avoid this error.

src/template_finder.py
mgleed: To clarify for @Ezro and @aeon0 

`search_all()` is for finding all matches from a list of templates on screen. It returns a list of match objects. It iterates over the cv2.matchTemplate result and continuously masks it until there's no results left. So I input ["quarter", "dime"] and the picture were a picture of a ton of coins, it'd result a list of all the quarter and dime matches in the image. 

`search_all()` works, but it's not optimal. the optimal method is as `_alex_` described in that discord link. 

`search_all_templates()` returns the first match of all templates of interest on the screen. So if I input ["quarter", "dime"] and the picture were a picture of a ton of coins, it'd result the first match of quarter and the first match of dime. 

Perhaps there's a way we can better name these functions? Or restructure template_finder somehow to accommodate?

aeon0: But esentially the two functions do more or less very similar things and also share the same code structure. Couldnt we do it with a flag?

src/template_finder.py
Ezro: stored_templates_by_dir and search_all_templates have been removed

src/template_finder.py
Ezro: Resolving this as the PR has been updated to use a flag in search_all

mgleed:
@Ezro left a few comments earlier

aeon0:
From my side it looks good. I leave it to gleed since he seems to still have some topics.

mgleed:
Also, why'd you choose to stick with precise keyfile name rather than use regexp with character name like we discussed?

src/bot.py
```diff
@@ -257,19 +261,28 @@ def on_create_game(self):
...
             belt.fill_up_belt_from_inventory(Config().char["num_loot_columns"])
+        if not self._hotkeys_discovered:
+            saved_games_folder = Config().general["saved_games_folder"]
+            key_file = Config().general["key_file"]
+            hotkeys.discover_hotkey_mappings(saved_games_folder, key_file)
+            self._hotkeys_discovered = True
         self._char.discover_capabilities()
-        if corpse_present and self._char.capabilities.can_teleport_with_charges and not self._char.select_tp():
-            keybind = Config().char["teleport"]
+        teleport_selected = skills.select_tp(hotkeys.right_skill_key_map[SkillName.Teleport])
```
mgleed, line 276: Why do you need to pass `hotkeys.right_skill_key_map[SkillName.Teleport]` as an argument? Can't you just use bake this hotkey into `select_tp()`?

src/inventory/belt.py
```diff
@@ -25,7 +27,7 @@ def open(img: np.ndarray = None) -> np.ndarray:
 def close(img: np.ndarray = None) -> np.ndarray:
     img = grab() if img is None else img
     if not is_visible(ScreenObjects.BeltExpandable, img):
-        keyboard.send("esc")
+        keyboard.send(hotkeys.d2r_keymap[hotkeys.HotkeyName.OpenMenu])
```
mgleed, line 30: This place and others might be optimal to use new hotkey for ClearScreen. Though this may require some logic since I dont think ClearScreen is a required hotkey. Not necessarily in this PR so we can avoid regression, just thinking out loud.

src/pather.py
```diff
@@ -501,7 +503,7 @@ def _convert_rel_to_abs(rel_loc: tuple[float, float], pos_abs: tuple[float, floa
         return (rel_loc[0] + pos_abs[0], rel_loc[1] + pos_abs[1])
 
     def traverse_nodes_fixed(self, key: str | list[tuple[float, float]], char: IChar) -> bool:
-        if not char.capabilities.can_teleport_natively:
+        if SkillName.Teleport not in hotkeys.right_skill_key_map:
```
mgleed, line 506: Have you tested this behavior with charged amulet users? Fixed nodes require teleport in that they sometimes traverse walls. What happens if charged user runs out of charges during traverse? If can_teleport_natively works here still, then I'd leave this piece alone for this PR.

src/utils/hotkeys.py
```diff
@@ -0,0 +1,342 @@
...
+    PartyOnAutomap = 'PartyOnAutomap',
+    NamesOnAutomap = 'NamesOnAutomap',
+    ToggleMiniMap = 'ToggleMiniMap',
+    SayHelp = 'SayHelp',
+    SayFollowMe = 'SayFollowMe',
+    SayThisIsForYou = 'SayThisIsForYou',
+    SayThanks = 'SayThanks',
+    SaySorry = 'SaySorry',
+    SayBye = 'SayBye',
+    SayNowYouDie = 'SayNowYouDie',
```
mgleed, line 84: Lol, I'm just imagining integrating these in the runs.

```py
keyboard.send(hotkey.SayNowYouDie)
kill_diablo()
```

mgleed (dismissed):
Overall very well done! See comments. Thanks for doing this.

Last wish list item for this PR would be basic validation.

There are some hotkeys that simply cannot be unset. Like force move, etc. Do you think it'd be hard to implement some basic validation for required hotkeys?

src/inventory/belt.py
Ezro: I was thinking about that too but that's why I opted to use OpenMenu since it's a hidden hotkey defaulted to be "Esc" (and is not actually bindable in-game); it may be worth it to have a convenience somehow where the logic uses ClearScreen if ClearScreen not None else OpenMenu.

src/template_finder.py
```diff
@@ -211,55 +211,35 @@ def search_and_wait(
...
     color_match: list = False,
+    first_match: list = False
 ) -> list[TemplateMatch]:
     """
     Returns a list of all templates scoring above set threshold on the screen
     :Other params are the same as for template_finder.search()
     :return: Returns a list of TemplateMatch objects
     """
-    templates = _process_template_refs(ref)
+    templates = ref if first_match else _process_template_refs(ref)
```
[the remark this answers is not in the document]
Ezro: I've updated the logic to skip _process_template_refs when it's not a list[Template]

src/template_finder.py
```diff
@@ -211,55 +211,35 @@ def search_and_wait(
...
 
 def search_all(
-    ref: str | np.ndarray | list[str],
+    ref: str | np.ndarray | list[str] | list[Template],
     inp_img: np.ndarray,
     threshold: float = 0.68,
     roi: list[float] = None,
     use_grayscale: bool = False,
     color_match: list = False,
+    first_match: list = False
```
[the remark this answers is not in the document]
Ezro: first_match updated to best_match

src/bot.py
Ezro: We could do that. I think that's beyond the scope of this, though, because that'd be refactoring the seams of the code.

That's to say that prior to my change we passed in Config().char["teleport"] and my change aims to hotswap that out with hotkeys.right_skill_key_map[SkillName.Teleport]

Ezro:
> 



> Overall very well done! See comments. Thanks for doing this.
> 
> Last wish list item for this PR would be basic validation.
> 
> There are some hotkeys that simply cannot be unset. Like force move, etc. Do you think it'd be hard to implement some basic validation for required hotkeys?

I think we can implement that very easily (at some point). [Here's the current spot for parsing the key block into actual hotkeys](https://github.com/Ezro/botty/blob/feature/auto_detect_hotkeys/src/utils/hotkeys.py#L258)

src/template_finder.py
Ezro: @mgleed I added some asserts but they're very basic; please validate

mgleed:
> > Also, why'd you choose to stick with precise keyfile name rather than use regexp with character name like we discussed?
> 
> I haven't tested this but I think it's better to be explicit; I have a feeling that the hash is there for multi-account handling; e.g., if account A has character 'Paladin' and account B has character 'Paladin' and both were played online on the same PC then we wouldn't actually be able to determine which Paladin file to use.
> 
> This adds a bit of overhead for the users (to find and enter the explicit key file name) but I think the trade-off is that Botty wouldn't have to do decision making logic for choosing / erroring.

That's a good point and I'd bet your theory is correct.

mgleed (dismissed):
Testing locally. if it works well I'll merge today :tm:

mgleed:
Are you able to recover my test that I wrote in template_finder_test.py for this? Looks like that got overwritten with a force push.

Issue #910, opened 2022-06-20
Not stashing gold (Gold Barb Trav)


duarte3030:
**Is it really a Bug report?**
Yes.

**Botty Version**
0.8.0 - Release from GitHub

**Describe the bug**
On Trav runs only, Barb build nightmare:
1. After getting the last bit of gold and becoming full, it successfully Flags true for Full Gold.
2. On the following run, it buys potions before, spends gold, then the flag shows as False right before it's supposed to go to stash.

**To Reproduce**
1. Barbarian Travincal Nightmare runs, runs_per_stash = 1, stash_gold = 1, sell_junk = 0.
2. Pickup the last bit of gold to become full.
3. Next run, it goes to Ormus first, spends money on potions, and doesn't stash gold (flag becomes false for full gold).

**Expected behavior**
Either stash first, or stash on 98% full and above for example (: 

**Screenshots / Video**

**Logs**

Actually it's downright skipping the stash:

[0.8.0 2022-06-20 02:36:38,819] DEBUG      Pick up expression: [type] == gold # [gold] >= 5000
[0.8.0 2022-06-20 02:36:38,821] INFO       Attempt to pick up 15215 GOLD
[0.8.0 2022-06-20 02:36:39,788] DEBUG      Read 16 ground items in 0.503 seconds
[0.8.0 2022-06-20 02:36:39,804] INFO       Set inventory gold full: True
[0.8.0 2022-06-20 02:36:41,020] INFO       Health Manager is now paused
[0.8.0 2022-06-20 02:36:41,022] INFO       End game. Elapsed time: 81.24s
[0.8.0 2022-06-20 02:36:41,031] INFO       Starting game #4
[0.8.0 2022-06-20 02:36:41,032] DEBUG      Waiting for templates: ['MAIN_MENU_TOP_LEFT', 'MAIN_MENU_TOP_LEFT_DARK', 'A5_TOWN_0', 'A5_TOWN_1', 'A4_TOWN_4', 'A4_TOWN_5', 'A3_TOWN_0', 'A3_TOWN_1', 'A2_TOWN_0', 'A2_TOWN_1', 'A2_TOWN_10', 'A1_TOWN_1', 'A1_TOWN_3']
[0.8.0 2022-06-20 02:36:44,149] DEBUG      Found match: MAIN_MENU_TOP_LEFT (96.9% confidence)
[0.8.0 2022-06-20 02:36:45,271] DEBUG      Saved character template found and already highlighted, continue
[...]

PR #922, opened 2022-06-22
Fix not stashing gold if buying sth before

aeon0:
fixing: #910

### DIFF
--- a/src/bot.py
+++ b/src/bot.py
@@ -329,6 +329,7 @@ class Bot:
                 items = personal.inspect_items(game_stats=self._game_stats)
         keep_items = any([item.keep for item in items]) if items else None
         sell_items = any([item.sell for item in items]) if items else None
+        stash_gold = personal.get_inventory_gold_full()
 
         # Check if should need some healing
         img = grab()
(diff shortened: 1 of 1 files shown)


src/bot.py
```diff
@@ -329,6 +329,7 @@ def on_maintenance(self):
                 items = personal.inspect_items(game_stats=self._game_stats)
         keep_items = any([item.keep for item in items]) if items else None
         sell_items = any([item.sell for item in items]) if items else None
+        stash_gold = personal.get_inventory_gold_full()
```
aeon0, line 332: the inventory full of gold flag is overwritte if we buy any potions. And technically, it is right. We are not full of gold anymore... But we probably still want to stash gold : )

roozhou:
Should reset gold flag after repair and resurrect as well.

aeon0:
> Should reset gold flag after repair and resurrect as well.

Question is, do we really want to reset at all these places. If your char is full of gold you have close to e.g. 900k (assuming lvl 90). If you ressurect for 50k or buy some positions you probably still want to stash or have the full of gold flag set.

PR #926, opened 2022-06-22
Replace nip with bnip

Riddle1001:


mgleed:
This looks fine. Will you run your bot for 30 mins or so with this PR and let me know if it has any issues?

PR #928, opened 2022-06-23
Update diablo.py _sealdance()

xw220:
I watched 10 hours of diablo run by 4 fohdins. About 30% were failed games which aborted in middle run, especially on detecting that if the seal is openned.  In almost every fail caused by this reason, the seal was openned successfully, but the decisions were closed for 5 times then the game was aborted. 

By increasing the numbers of try and reduce the seal_opentemplates threshold to 0.7, the incorrect judgements on seal's open status seldom occur. Now the ratio of failed games typically reduced. It is about 12% mostly caused by temporary high latency or getting lost after teleport to pick item.

### DIFF
--- a/src/run/diablo.py
+++ b/src/run/diablo.py
@@ -113,11 +113,11 @@ class Diablo:
     # OPEN SEALS
     def _sealdance(self, seal_opentemplates: list[str], seal_closedtemplates: list[str], seal_layout: str, seal_node: str) -> bool:
         i = 0
-        while i < 4:
+        while i < 8:
             Logger.debug(seal_layout + ": trying to open (try #" + str(i+1)+")")
             self._char.select_by_template(seal_closedtemplates, threshold=0.5, timeout=0.1, telekinesis=True)
             wait(i*0.5)
-            found = template_finder.search_and_wait(seal_opentemplates, threshold=0.75, timeout=0.1).valid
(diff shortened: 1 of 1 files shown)


Riddle1001:
@xw220 There are a couple other people working on chaos sanctuary too, if you'd like to talk to them join us on discord and @ one of the admins to get the contributor role. https://discord.gg/Jf3J8cuXWg

jobithu:
Interesting, because with the original parameters, the pressure tests revealed >95% success rate.

If the reduction from 0.75 to 0.75 does not cause any false-postives (closed seals considered open) I wouldn't mind changing this value. 

When it comes to number of tries: i had 7 when i started, then lowered to 5 to save time at runs. typically if you failed to open 5 times already, then something is really wrong, so adding another 2 or 3 tries most of the times wont change it and its better to accept that the run was a fail and keep going with a new one.

by the way: we are currently completely reworking diablo. you might be interested looking into that PR (-> https://github.com/bottytools/botty/pull/867). also, we consider to remove the template detection for seals and use hard-coded clicks to see if it is more efficient.

xw220:
> Most important point: do you have REDEMPTION bound to skills? typically templates for checking seals are obscured by corpses that are close by. redemption removes these to increase template matching.

Yes, I bound redemption.

> If the reduction from 0.75 to 0.75 does not cause any false-postives (closed seals considered open) I wouldn't mind changing this value.

Reduce to 0.7 do helps. I saw many 72% 73% in debug logs.

PR #944, opened 2022-06-25
New Hydra Frozen Orb Build

Arcia125:
Only runs Pindle, Eldritch, and Shenk. Similar to Hydra sorc, but utilizes both Frozen Orb and FIreball.


D2RLegit:
Unfortunately all new chars will have to wait and will need to be restructered after a couple overhauls have been done shortly. Thanks for the pr for the time being though.

